Issue #131, opened 2019-03-08
Learn the CSS Box Model by Building a Rothko Painting


QuincyLarson:

scissorsneedfoodtoo:
@QuincyLarson, is it okay if I start working on this?

scissorsneedfoodtoo:
Here's an example project: https://codepen.io/scissorsneedfoodtoo/pen/KEoKqe

QuincyLarson:
@scissorsneedfoodtoo Well done! That looks great!

Can you work backward to break the building of the project down into steps like these?

https://github.com/freeCodeCamp/CurriculumExpansion/pull/135/files#diff-99314d694942b489d28d8a6f0c3bc5a3

scissorsneedfoodtoo:
Sure! I'll use that as an inspiration to break the project down.

PR #138, opened 2019-04-01
feat: first draft of Rothko CSS project instructions

scissorsneedfoodtoo:
Rough draft of the Rothko project.

For issue https://github.com/freeCodeCamp/CurriculumExpansion/issues/131


Ethan-Arrowood:
I wonder if it would be better to put the challenge text in the CSS file itself rather than the HTML considering that is where a majority of the code changes are happening. We can make the HTML changes for the students automatically and let them only have to worry about editing the CSS file.

scissorsneedfoodtoo:
@Ethan-Arrowood, yes, very good point! I'll go through the files and fix that.

scissorsneedfoodtoo:
@beaucarnes, thank you for your review! Yes, that's a very good point. I'll look into adding both a description and diagram of the box model at the beginning of the project.

scissorsneedfoodtoo:
@Ethan-Arrowood and @beaucarnes, sorry, my earlier push was rejected because I should have force pushed instead. The new diagrams and descriptions should be there now. Please take a look when you can!

beaucarnes:
Interesting method to have users put the img url in to see the image. I like it.

scissorsneedfoodtoo:
Thanks! I thought about building it out in HTML and CSS, but thought it would be misleading since you can't set apply background color to padding.

scissorsneedfoodtoo:
Will merge this now just to clear the docket. Please leave a comment in #131 if you notice any issues!

Issue #113, opened 2019-03-08
Learn JSON by Building a Sports Team


QuincyLarson:

HKuz:
Here are my initial thoughts for the project - I made a sports team vs a league, but can tweak this if we want. I went for an international, retro flair with the Argentinian World Cup winning team in 1986, mainly because I think the story that one of their players (Diego Maradona) hand-balled a goal in the tournament and called it the 'hand of god' is funny 🤣 .

# JSON Project

## Concepts to teach

- JavaScript Object Notation (JSON) is a text file format to structure data
- Popular structure to exchange data over the web because: lightweight, easily exchanged, human readable, easily parsed by machines
- JSON started as a JavaScript standard, but it's language independent and can be used in many programming languages
- Build JSON using two structures:
    1. **Objects:** an unordered collection of name-value pairs
    2. **Arrays:** an ordered collection of values
- Values can be:
    - Numbers
    - Boolean values
    - Strings (must use double quotes)
    - Objects
    - Arrays
    - null
- How to access data in JSON format
    - Dot notation (objects only, uses name but it can't have spaces)
    - Bracket notation (arrays use index, object uses name)

## Data

The example structure below would make for a lot of typing for campers in its entirety, so I see them building it up only in the beginning. They can add the first several fields with different data types, then 1-2 entries in the players array. The project can end with some practice accessing different items.

- **Data Source:** [Wikipedia](https://en.wikipedia.org/wiki/1986_FIFA_World_Cup_squads#Argentina)

HKuz:
First draft of the JSON project:

```js
{
    "Team": "Argentina",
    "Sport": "Football",
    "Year": 1986,
    "World Cup Winner": true,
    "Head Coach": {
        "Name": "Carlos Bilardo",
        "Matches": 7
    },
    "Players": [
        {
            "Name": "Sergio Almirón",
            "Position": "FW",
            "Number": 1,
            "isCaptain": false,
            "Nickname": null
        },
        {
            "Name": "Sergio Batista",
            "Position": "MF",
            "Number": 2,
            "isCaptain": false,
            "Nickname": null
        },
        {
            "Name": "Ricardo Bochini",
            "Position": "MF",
            "Number": 3,
            "isCaptain": false,
            "Nickname": "El Bocha"
        },
        {
            "Name": "Claudio Borghi",
            "Position": "MF",
            "Number": 4,
            "isCaptain": false,
            "Nickname": "Bichi"
        },
[...]

PR #140, opened 2019-04-03
feat: first draft of the JSON curriculum

HKuz:
JSON challenges for issue #113 

Challenges are meant to address:
- valid JSON syntax
- stringifying an object to transfer over the web
- parsing a JSON string into the object it describes
- practice accessing values at various levels in an object, including looping over arrays within a JSON structure

Full code is saved in `index.js`.

Notes for reviewers - feedback is welcome on these!:
- Some of the challenges that explain concepts are a bit verbose - I think they still stay within the 2 minute rule, but if there are ways to trim them, suggestions welcome
- I have campers build the JSON object in a `.js` file (vs a `.json` file) so they can practice stringifying, parsing, and accessing it later in the lessons
- I'm rusty using JavaScript, let me know if you notice any conventions in the code that wouldn't be considered 'community accepted' anymore

### DIFF
--- a/data-visualization/json/index01.js
+++ b/data-visualization/json/index01.js
@@ -1,0 +1,15 @@
+
+
+/*
+JavaScript Object Notation (JSON) is a text format that you use to structure data.
+
+Because JSON is lightweight, human-readable, and easily parsed by machines, it's a popular structure to exchange data over the web.
+
+It started as a JavaScript standard, but now it is language independent and can be used in many programming languages.
+
+JSON is built using two structures:
+1. Objects: unordered collections of key-value pairs, surrounded by `{ }`, referred to as curly braces
+2. Arrays: ordered collections of values, surrounded by `[ ]`, referred to as square braces
+
+Declare a new variable using the `const` keyword and name it `myTeam`, then set it equal to an empty object.
+*/
--- a/data-visualization/json/index02.js
+++ b/data-visualization/json/index02.js
@@ -1,0 +1,21 @@
+const myTeam = {};
+
+/*
+The last challenge mentioned that you structure your data in JSON with objects and arrays.
+
+Both of those structures contain values, which can be of any of the following types:
+- Strings (must use double quotes)
+- Numbers
+- Boolean values (`true` or `false`)
+- Objects
+- Arrays
+- `null`
+
+Note that JavaScript object literal syntax lets you type a key without any quotes around it. This is one place where JSON differs - make sure to put quotes around all key names.
+
+Also, the JSON specification requires you to surround any string with double quotes (`" "`), not single quotes (`' '`).
+
+You're going to build a structure in JSON to hold the 1986 World Cup champion team from Argentina.
+
+Start by adding a new key into your `myTeam` object called `Team` with a string value of `Argentina`.
+*/
--- a/data-visualization/json/index03.js
+++ b/data-visualization/json/index03.js
@@ -1,0 +1,9 @@
+const myTeam = {
+  "Team": "Argentina"
+};
+
+/*
+Now add two more key-value pairs to your `myTeam` object. One for `Sport` with a string value of `Football` (remember to use double quotes `"`), and the other for `Year` with a numeric value of `1986` (remember not to use any quotation marks).
+
+Remember that key-value pairs in objects are separated by commas.
+*/
--- a/data-visualization/json/index04.js
+++ b/data-visualization/json/index04.js
@@ -1,0 +1,15 @@
+const myTeam = {
+  "Team": "Argentina",
+  "Sport": "Football",
+  "Year": 1986
+};
+
+/*
+Key names that include spaces are valid JSON.
+
+A quick aside about JavaScript syntax and JSON. Recall that JavaScript object literal syntax lets you type a key without any quotes around it. This only works when your key name is a valid JavaScript identifier: it doesn't have spaces, hyphens, or start with a number. If you want to use these characters in a key name when you're creating a JavaScript object, you would need to put the key name in quotes - does this sound familiar?
+
+JSON already requires that you surround all string values with double quotes. Because of this, JSON is very flexible with what you're allowed to use as valid key names to structure your data.
+
+Add another key-value pair after `Year` for `Is World Cup Winner`, this time use a Boolean value set to `true`. Remember not to surround this value with quotation marks - `"true"` is not the same as `true` - one is a string, the other is a Boolean value.
+*/
--- a/data-visualization/json/index08.js
+++ b/data-visualization/json/index08.js
@@ -1,0 +1,22 @@
+const myTeam = {
+  "Team": "Argentina",
+  "Sport": "Football",
+  "Year": 1986,
+  "Is World Cup Winner": true,
+  "Head Coach": {
+    "Name": "Carlos Bilardo",
+    "Matches": 7
+  },
+  "Players": [
+    {
+      "Name": "Sergio Batista",
+      "Position": "MF"
+    }
+  ]
+};
+
+/*
+Add two more key-value pairs to Sergio's object: `Number` set to a numeric value of `2`, and `isCaptain` set to a Boolean value of `false`.
+
+Note the `is` part of `isCaptain`, and earlier you saw `Is` in the key name `Is World Cup Winner`. This makes the key name sound like a yes/no (or true/false) question. A common convention you'll see when a variable or key is set to a Boolean value is to start the name with `is`. This indicates to anyone reading your code or accessing that key to expect a Boolean value.
+*/
--- a/data-visualization/json/index10.js
+++ b/data-visualization/json/index10.js
@@ -1,0 +1,29 @@
+const myTeam = {
+  "Team": "Argentina",
+  "Sport": "Football",
+  "Year": 1986,
+  "Is World Cup Winner": true,
+  "Head Coach": {
+    "Name": "Carlos Bilardo",
+    "Matches": 7
+  },
+  "Players": [
+    {
+      "Name": "Sergio Batista",
+      "Position": "MF",
+      "Number": 2,
+      "isCaptain": false,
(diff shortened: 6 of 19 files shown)


data-visualization/json/index01.js
```diff
@@ -0,0 +1,15 @@
...
+
+/*
+JavaScript Object Notation (JSON) is a text format that you use to structure data.
+
+Because JSON is lightweight, human readable, and easily parsed by machines, it's a popular structure to exchange data over the web.
+
+It started as a JavaScript standard, but now it is language independent and can be used in many programming languages.
+
+JSON is built using two structures:
+1. Objects: unordered collections of key-value pairs, surrounded by `{ }`
```
Ethan-Arrowood:
```suggestion
1. Objects: unordered collections of key-value pairs, surrounded by `{ }`, referred to as curly braces
```

data-visualization/json/index01.js
```diff
@@ -0,0 +1,15 @@
...
+2. Arrays: ordered collections of values, surrounded by `[ ]`
```
Ethan-Arrowood:
```suggestion
2. Arrays: ordered collections of values, surrounded by `[ ]`, referred to as square braces
```

data-visualization/json/index01.js
```diff
@@ -0,0 +1,15 @@
...
+
+Declare a new variable using the `let` keyword and name it `myTeam`, then set it equal to an empty object.
```
Ethan-Arrowood:
```suggestion
Declare a new variable using the `const` keyword and name it `myTeam`, then set it equal to an empty object.
```
Would rather use const if we aren't reassigning something to this variable

data-visualization/json/index02.js
```diff
@@ -0,0 +1,21 @@
+let myTeam = {};
```
Ethan-Arrowood:
```suggestion
const myTeam = {};
```
`const` key word change

From here on I wont suggest the `const` change -- make those changes locally and push them up (or if we decide to use `let` then ignore 😁)

data-visualization/json/index02.js
```diff
@@ -0,0 +1,21 @@
...
+
+/*
+The last challenge mentioned that you structure your data in JSON with objects and arrays.
+
+Both of those structures contain values, which can be the following types:
```
Ethan-Arrowood:
```suggestion
Both of those structures contain values, which can be of any of the following types:
```

data-visualization/json/index02.js
```diff
@@ -0,0 +1,21 @@
...
+- Strings (must use double quotes)
+- Numbers
+- Boolean values (`true` or `false`)
+- Objects
+- Arrays
+- `null`
+
+Note that JavaScript object literal syntax lets you type a key without any quotes around it. This is one place where JSON differs - make sure to put quotes around any key name.
```
Ethan-Arrowood:
```suggestion
Note that JavaScript object literal syntax lets you type a key without any quotes around it. This is one place where JSON differs - make sure to put quotes around all key names.
```

data-visualization/json/index03.js
```diff
@@ -0,0 +1,9 @@
+let myTeam = {
+  "Team": "Argentina"
+};
+
+/*
+Now add two more key-value pairs to your `myTeam` object. One for `Sport` with a value of `Football`, the other for `Year` with a value of `1986` (this should be a number, not a string).
```
Ethan-Arrowood:
```suggestion
Now add two more key-value pairs to your `myTeam` object. One for `Sport` with a string value of `Football` (remember to use double quotes `"`), and the other for `Year` with a number value of `1986` (remember not to use any quotation marks).
```

data-visualization/json/index04.js
```diff
@@ -0,0 +1,9 @@
+let myTeam = {
+  "Team": "Argentina",
+  "Sport": "Football",
+  "Year": 1986
+};
+
+/*
+Add another key-value pair after `Year` for `World Cup Winner`, this time use a Boolean value set to `true`.
```
Ethan-Arrowood: Add a line here about how properties can have spaces within them. This is unique to JSON (as compared to regular JS objects) so it would be good to cover that here.

Additionally:
```suggestion
Add another key-value pair after `Year` for `World Cup Winner`, this time use a Boolean value set to `true` (remember not to surround this value with quotation marks).
```

data-visualization/json/index08.js
```diff
@@ -0,0 +1,20 @@
...
+  "Players": [
+    {
+      "Name": "Sergio Batista",
+      "Position": "MF"
+    }
+  ]
+};
+
+/*
+Add two more key-value pairs to Sergio's object: `Number` set to a numeric value of `2`, and `isCaptain` set to a Boolean value of `false`.
```
Ethan-Arrowood: Add a sentence to describe the `isCaptain` notation. (like the lowercase of the word `is` and how it is commonly used to denote a boolean property)

data-visualization/json/index10.js
```diff
@@ -0,0 +1,29 @@
...
+    }
+  ]
+}
+
+/*
+Now you've seen an example of each valid data type in JSON: numbers, strings, Boolean values, objects, arrays, and null.
+
+JSON can look a lot like JavaScript object literal sytax, but there are two differences worth noting here:
+1. JSON is a data format only - it's used to hold properties, so including methods in an object is NOT valid JSON
+2. JSON doesn't need to be an object (`{ }`) - any of the data types mentioned above could be valid JSON on their own (a single string or number), but some JSON data will have an array as the outer-most structure
```
Ethan-Arrowood: I love this line! Would be great to restate both of these points at the very end of the challenge.

data-visualization/json/index12.js
```diff
@@ -0,0 +1,177 @@
...
+
+
+/*
+To save you some typing, we filled in the rest of the player array - notice that all the objects in it use the same key names.
+
+Normally, you would save data in valid JSON format to a text file with a `.json` extension, without the variable declaration or ending semicolon.
+
+Here, we're going to convert our JavaScript object into a valid JSON string ready for web transport through a process called "stringification". You would use the `JSON.stringify()` method, and pass it your object.
+
+Stringify `myTeam` and store it in a new variable called `myString` (declare your variable with the `let` keyword).
```
Ethan-Arrowood:
```suggestion
Stringify `myTeam` and store it in a new variable called `myString` (declare your variable with the `const` keyword).
```

data-visualization/json/index13.js
```diff
@@ -0,0 +1,181 @@
...
+/*
+Now you have a JSON string ready to send over the web! Typically, you won't type out your own JSON data, but instead you'll be working with it.
+
+Normally, you would send a request to retrieve the JSON data you want to work with (this concept is covered in the AJAX section). JSON data transferred over the web is in the stringified format.
+
+There's one more step you may need to do to work with stringified JSON data in your code - you need to parse it with the `JSON.parse()` method. This method takes the JSON string as an argument and recreates the JavaScript value or object that's described by the string. Note that this step may be done for you, depending on the way you requested the data - make sure to read the documentation!
+
+For simplicity, we'll pretend we received our strigified data (stored in `myString`) over the web.
+
+Parse `myString` and store it in a new variable called `myData` (declare your variable with the `let` keyword).
```
Ethan-Arrowood:
```suggestion
Parse `myString` and store it in a new variable called `myData` (declare your variable with the `const` keyword).
```

HKuz:
Great comments and suggestions, @Ethan-Arrowood - and thanks for the quick review! Just made the changes and added a final challenge - there's no coding part - it shows the 'answer' from the last challenge, but is really meant to be a recap of the major concepts.

RandellDawson:
If this lesson is about JSON, then they should be creating a JSON file vs. creating an object which with JSON syntax.  I don't know how you will do it, but I think the current way you are creating an object and assigning to the `myTeam` variable, then converting it to a string and then parsing that string could be very confusing to a user and it is not "real-life".  

The following should not even be necessary:
```
const myString = JSON.stringify(myTeam);
```
The user needs to create a JSON file (in some way), then you still use similar verbiage (you already have) about how this JSON would be pulled into via AJAX.  You could explain how using AJAX the JSON would be assigned to a variable (myTeam) and then it would make sense to talk about the `parse` method.

Ethan-Arrowood:
I disagree @RandellDawson -- often time JSON data structures are defined in js and are then stringified to be sent over HTTP. Challenge 13 does a good job of explaining the AJAX case your mention. Additionally, JSON is 'JavaScript Object Notation' so I'm very supportive of teaching it within a JavaScript context. Lastly, I do not believe comments are allowed in `.json` files which would make it very difficult to author the challenges

HKuz:
Hey guys, good discussion going on here - better to hash this out now vs when these are live! @Ethan-Arrowood makes some good points on the theory side - two major concepts I think these challenges were trying to get across are a) the specific JSON syntax, and b) since campers are learning JavaScript in tandem, to really highlight the differences as well as similarities with JS objects.  I think teaching them these within the JavaScript context does help, at least to really stress the latter concept. 

If we do go the `.json` file route, from a practical perspective I'd want to understand fCC's platform capabilities better - would we be able to request the data in the editor? Also, I know we're aiming to eventually make fCC available offline - may want to keep that as a consideration as we wire these challenges.

RandellDawson:
> often time JSON data structures are defined in js and are then stringified to be sent over HTTP.

That is fine, but converting a JavaScript object to a stringified object and then back to a JavaScript object seems redundant and you would never do that in a real project.  You might take a JSON retrieved from an JSON and `parse` it or you might take a JavaScript object and stringify it to send it over HTTP, but you would never do both.  If you can think of a practical example of both, then I suggest you incorporate it into the challenge in some way, though I think that is beyond the scope of the actual challenge IMHO.

Ethan-Arrowood:
I imagine those expectations would be better for the Node.js based challenges. Especially those dealing with networking

PR #144, opened 2019-04-10
fix: update html project tests

beaucarnes:
I moved the tests in the HTML project files to be near where the code should be written. I also added backticks around code in the tests.


scissorsneedfoodtoo:
@beaucarnes, thank you for bringing this project over and moving all the tests around. Everything LGTM!

Issue #130, opened 2019-03-08
Learn CSS Variables by Building a City Skyline


QuincyLarson:

ahmadabdolsaheb:
I removed my assignment just in case someone is interested in contributing to this while I work on the other parts of the curriculum.

QuincyLarson:
@ahmadabdolsaheb Sounds good. Thanks for freeing this up.

thecodingaviator:
@QuincyLarson Can I pick this one up? You can check my [CodePen](https://codepen.io/thecodingaviator/pens/public/) I have some examples of CSS things I've done over there, if it looks good, please assign it!

moT01:
I can try and take this one. So what's going to happen to flappy the penguin? ☹️

QuincyLarson:
@moT01 We may be able to keep it as a separate project, like "Learn Basic CSS by Building a CSS Penguin" or something like that.

moT01:
Looking for some initial feedback...
full page -> https://codepen.io/moT01/full/RdzNPr
code -> https://codepen.io/moT01/pen/RdzNPr?editors=1100

So are we going to want users to create all the code?
Or is it acceptable for me to like create some div's or something and just tell them to "add 'class' to all the div's"

ahmadabdolsaheb:
@moT01, so on this curriculum iteration we are focusing on repetition. So the project need to be built from the first line of code. you could checkout other topics for reference.

thecodingaviator:
@moT01 Looks awesome!

moT01:
Thanks guys! So will I want a single HTML file with all the CSS in style tags like the current challenges? And how much do I want in there? doctype, html, head? Have the user create all of that, or will the editor basically be the body tag and just have them create the stuff in there?

QuincyLarson:
@moT01 That's a thing of beauty. Brought a tear to my eye.

Yes - we will want to have people write ever line of code if possible. In doing so, not only do they get the repetition @ahmadabdolsaheb mentioned, but they also get more of a feel of accomplishment that they are actually building the full project themselves.

moT01:
Thanks @QuincyLarson, I used a lot of `linear-gradient` - I'm not sure how much that will be covered in other projects, should we expand this to cover that in depth a little more? Maybe `Learn CSS Variables and Linear-Gradient by Building a Skyline`. I will pretty much have to cover it regardless, but maybe not as much if it was already taught.

[...]

PR #145, opened 2019-04-12
feat/css-variables-first-draft

moT01:
Initial commit of CSS variables project. Any feedback is welcome.

issue #130


Ethan-Arrowood:
Wow 100 challenges?! Great work; this is going to take some serious time to review.

moT01:
Yea, I got a little worried about the length as I was getting deep into it. I would have made a few less buildings if I could start over, but it's really difficult to go back and make a large change like that.

scissorsneedfoodtoo:
@moT01, took a quick look at your draft here and it's looking great! As you mentioned, you could simplify the project a bit by removing some buildings, but I understand how time consuming it can be to go back and change all of these files. It shouldn't be a problem so long as the broken down project is less than 200 tests.

I just noticed a couple of things. First, the class names could be more descriptive. I know that they're abbreviations for background building or foreground building and their sections, but it might be better to teach users to favor longer, easier to read class names. This is a pretty minor detail for now and can be changed later. 

Next, each challenge will only test one thing, so files like `index030.html` could be broken down into two steps: one for using the `*` and applying a border, and another for applying `box-sizing: border-box;`. You could also go back and break adding the doctype, head, html, and body tags into different steps, but I don't think it's really necessary. Starting with the skeletal HTML page and having the user add the title as a first step is fine, too.

moT01:
Thanks for the feedback @scissorsneedfoodtoo. So are you kind of saying that most of the challenges need to be broken up? 

Like `index050`, for instance, should be three challenges? One to create a div with the class, one for adding a width, and one for adding a height?

I pretty much did some stuff like that on every challenge that could be broken up. I thought that someone who has done the basic html and css should be able to pretty easily add elements and give them a few small properties. Breaking all those up would lead to many repetitive width and height challenges and would definitely go beyond the 200.

scissorsneedfoodtoo:
@moT01, I'm sorry, I could have explained this better. You make a good point that the users probably would have finished the basic HTML and CSS projects by the time they're working on this one. Though it's my understanding that we'll want to break things as much as possible, you're right in thinking it would be tedious after a certain point.

Looking at the RPG game as an example, from `index010` to `index050` users declare variables one at a time, then in `index060` declare two variables. And in `index120`, users are asked to declare nine different variables using `querySelector()`. I suppose it would be more accurate to say that we should break things down into small steps when first introducing something, then expect more as users are familiar with what you want them to do.

I think for your project it would be good to start a bit slower as users build the first building, then quickly ramp up since the other buildings share a similar foundation (no pun intended). For example, in `index050` it might be good to have users create the `div` element and give it a class. Then in the next lesson (`index051`?), users can target the `background-buildings` class and give it a width and height of 100%. I could also see breaking `index060` into two steps, first to nest the `div` and give it a class, then to set its width and height. Once you get to `index070` and `index080`, they should know what to do, so there's no need to break those lessons down. Also, since you teach very useful techniques in `index030` and `index040` that have a big impact on the look of the page, so it might be good to break those down into two or three lessons apiece.

Hope that helps! You have a really cool project here and I can't wait to build a city skyline myself.

scissorsneedfoodtoo:
@moT01, just wanted to check in and see how everything's going. If you don't mind, I'll merge this and you can submit another PR to continue your work.

moT01:
I haven't really done anything since the PR. Yes, you can go ahead and merge it if you want and I'll just make another PR.

scissorsneedfoodtoo:
Great! I'll do that now.

PR #148, opened 2019-04-12
First draft of node-js-web-server

padunk:
This is the first draft of node js web server for fcc curriculum expansion

### DIFF
--- a/node-js-web-server/index010.js
+++ b/node-js-web-server/index010.js
@@ -1,0 +1,14 @@
+const http = require('http');
+
+// In order to create your first server, we need a http package.
+// require it in your index.js
+// Node http package have a standard method to build a server.
+// The method is called createServer. createServer takes an options
+// which is an Object and a requestListener, which is a function and returns
+// a http.Server instance
+// The syntax looks like this:
+// http.createServer(function(request, response) {
+//   // do some logic here
+// });
+// Now create your createServer and console.log the request url and method
+// using request.url and request.method.
--- a/node-js-web-server/index030.js
+++ b/node-js-web-server/index030.js
@@ -1,0 +1,19 @@
+const http = require('http');
+const fs = require('fs');
+
+http.createServer(function (req, res) {
+   console.log(req.url, req.method);
+
+});
+
+// Next to read the file you could use readFile method or createReadStream method from fs module.
+// you can use readFile method since the file is not big in size.
+// if you have file with big size, use createReadStream instead.
+// createFile will take a path, an options and a callback.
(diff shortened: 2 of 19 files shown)


scissorsneedfoodtoo:
@padunk, so sorry for the delay in reviewing this. I'll leave some suggestions as review comments.

node-js-web-server/index010.js
```diff
@@ -0,0 +1,14 @@
+const http = require('http');
+
+// In order to create your first server, we need a http package.
```
scissorsneedfoodtoo: This could be broken down more:

* Lesson 1:
"After including a module, you can use any of the methods attached to it. For example, `http.Agent();` uses the `Agent` method. Use the `http` module's `createServer` method to build a server."

* Lesson 2:
"The server is running, but won't do anything if it gets a request. To deal with any requests to the server, first pass a callback function to `createServer`. Here's an example of passing a callback function to `setTimeout`:
```
setTimeout(function() {

});
```

* Lesson 3:
"The callback function in `createServer` takes two parameters, `request` and `response`. Pass the two parameters to the callback function like so:
```
http.createServer(function(req, res) {

});
```

* Lesson 4, 5, and beyond:
I don't think it's necessary to have the student log req.url and req.method to the console. Instead, they could write something like this over a few lessons: 
```
if (req.url === '/') {
  res.write('Hello World');
  res.end();
}
```

Then we can create a simple page to show the "Hello World" message so that it's available offline. That, or students could just write `res.end("Hello World")` and see the simple page.

node-js-web-server/index030.js
```diff
@@ -0,0 +1,19 @@
+const http = require('http');
+const fs = require('fs');
+
+http.createServer(function (req, res) {
+   console.log(req.url, req.method);
+
+});
+
+// Next to read the file you could use readFile method or createReadStream method from fs module.
```
scissorsneedfoodtoo: This will also have to be broken down into multiple steps like above. You could briefly explain that simple pages have already been written as HTML files, they just need to be served when the user visits a certain URL.

Just go through step-by-step, first with `fs.readFile();`, then build up by having the student add `__dirname`, then `__dirname + '/static/hello.html'`, passing in the callback function, and so on. I know that just running `fs.readFile();` so having them start with writing `fs.readFile('', function() {});` could work, too.

scissorsneedfoodtoo:
I wasn't able to go through every lesson, but I'm sure you understand how much we want to break each lesson down. But your code is solid and I'm happy to merge this draft as it is! Go ahead and break down the lessons more, and submit another PR with a second draft whenever you can.

PR #149, opened 2019-04-13
fix: remove css-media-queries directory

scissorsneedfoodtoo:
Removed css-media-queries directory that was accidentally pushed to master and should have been submitted as a PR.


scissorsneedfoodtoo:
@Ethan-Arrowood, yes, that would be great! I don't seem to have the settings tab when I open the repo. Will have to check on the admin permissions for this repo.

PR #150, opened 2019-04-14
fix: accurately explain the differences between var and let

lionel-rowe:
If declared at the top level, `var` and `let` _both_ create global variables. The difference is that within a block that isn't a function body, `let` will be in the block scope, whereas `var` will be in the parent scope.

As we're at the top level here, the explanation needs to be changed to make sense.

Not sure if my suggested edit is all that informative --- perhaps could be improved further by adding more info or a link? Is linking to external resources encouraged or not?

### DIFF
--- a/learn-basic-javascript-by-building-a-role-playing-game/index040.js
+++ b/learn-basic-javascript-by-building-a-role-playing-game/index040.js
@@ -5,7 +5,7 @@ var currentWeapon = 0;
 
 /*
 We've been declaring variables with the 'var' keyword.
-This makes a variable available everywhere.
-But it's better to scope variable as narrowly as possible.
-Change every 'var' in the code to 'let' so that the variables have local scope instead of global scope.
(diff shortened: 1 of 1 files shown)


scissorsneedfoodtoo:
@lionel-rowe, thank you for bringing this up. You make a good point about both `var` and `let` creating global variables at this point.

The test description here could be tweaked to reflect that, or to give a different explanation of why using `let` is preferable to `var`. Either way, we should try to keep the explanation as short as possible, without over explaining or linking to external resources. Maybe it would be enough to write that using `var` can lead to strange behaviors, and that it's better to use the `let` keyword when declaring variables. @beaucarnes explains `const` and how it's used for variables that won't change in `index110.js`, so we don't need to go over that in this file.

beaucarnes:
@lionel-rowe You're right. Thanks for the suggestions. I simplified you're text a bit more.

Issue #122, opened 2019-03-08
Learn Regular Expressions by Building a Spam Filter


QuincyLarson:

lionel-rowe:
I'd be interested in contributing to this challenge.

QuincyLarson:
@lionel-rowe Awesome! I think this will be a fun one. See if you can get an extremely simple demo that checks a line of input for spam words and the use of characters meant to mask spam words like v1agr@ or something like that 😄 - I'm excited to see what you come up with.

lionel-rowe:
I'm actually wondering how best to approach this one. In order to make a truly extensible filter that would catch v1agr@, vi@gra, v|agRa, V I A G R A, etc. it'd really make more sense to have programmatically-generated regexes, rather than regex literals:

https://gist.github.com/lionel-rowe/0724546e4b5c1a71be29502aac4c825e

However, this approach would probably introduce too many disparate concepts at once, plus it has some gotchas (double-escaping, for starters). Any thoughts on how to simplify it while still guiding students toward writing extensible code?

Note that the regexes generated by this method are almost completely unreadable:
```js
[
  /\bv\s*[i|]\s*[a@4]\s*[g69]\s*r\s*[a@4]\b/i,
  /\bf\s*r\s*[e3]\s*[e3]\s* \s*m\s*[o0]\s*n\s*[e3]\s*y\b/i,
  /\bw\s*[o0]\s*r\s*k\s* \s*f\s*r\s*[o0]\s*m\s* \s*h\s*[o0]\s*m\s*[e3]\b/i,
  /\b[s5]\s*[t7]\s*[o0]\s*[c{\[(]\s*k\s* \s*[a@4]\s*l\s*[e3]\s*r\s*[t7]\b/i,
  /\bd\s*[e3]\s*[a@4]\s*r\s* \s*f\s*r\s*[i|]\s*[e3]\s*n\s*d\b/i
]
```

Alternatively, we could just take the hard-coding approach and limit the test cases somewhat to avoid having completely unreadable regexes.

brandenbyers:
Yes, programmatically might make the most sense in the real world but as you have pointed out, it really readable (nor does it teach much about regex). What about making the lesson slowly build on itself? Go from a normal spelling of a spam word. Next lesson is to add an alternative spelling. Next is more alternative spellings. Oh, wait...the spammers have gotten smarter and are now doing “x” things instead...how can we improve it from here?

[...]

PR #151, opened 2019-04-14
feat: first draft of Learn Regular Expressions by Building a Spam Filter

lionel-rowe:
WIP. Corresponds to issue #122. Early feedback welcome, especially regarding the instructions (tone of voice, verbosity, how easy they are to follow, etc.)

### DIFF
--- a/learn-regular-expressions-by-building-a-spam-filter/spamFilter010.js
+++ b/learn-regular-expressions-by-building-a-spam-filter/spamFilter010.js
@@ -1,0 +1,16 @@
+const isSpam = msg => {
+  /*
+
+  To start with, declare a variable called `re` using `const` and set its value to `/./`.
+
+  `/./` breaks down as follows:
+  * The starting and ending forward slashes (`/`) signal that it is a regex.
+  * The dot (`.`) means "match any character (other than new lines)".
+
+  Our regex will match any string that contains a character other than a new line. For example, "A", "嗡", and "asdf" all match, but "" doesn't.
+
+  Note that the match can be anywhere in the string.
+  */
+
+  return false;
+};
--- a/learn-regular-expressions-by-building-a-spam-filter/spamFilter040.js
+++ b/learn-regular-expressions-by-building-a-spam-filter/spamFilter040.js
@@ -1,0 +1,16 @@
+const isSpam = msg => {
+  const re = /viagra|free money|work from home|stock alert|dear friend/;
+  /*
+  Looking good!
+
+  However, the spammers are getting smarter. They've learned to sometimes mix up the case of their key phrases. As well as "free money", our filter also needs to be able to handle "FREE MONEY", "free MONEY", or even "fReE MoNEy".
+
+  Luckily, regex has just the tool for that: the case-insensitivity flag, `i`.
+
+  Regex flags are letters that you attach to the end of your regex to modify its behavior. For example, to create a case-insensitive regex for "foo bar", you would simply write `/foo bar/i`.
+
+  Add the `i` flag to the end of the regex.
+  */
+
+  return re.test(msg);
+};
--- a/learn-regular-expressions-by-building-a-spam-filter/spamFilter060.js
+++ b/learn-regular-expressions-by-building-a-spam-filter/spamFilter060.js
@@ -1,0 +1,27 @@
+const isSpam = msg => {
+  const blacklistRegexes = [
+    /viagra/i,
+    /free money/i,
+    /work from home/i,
+    /stock alert/i,
+    /dear friend/i
+  ];
+
+  // return re.test(msg);
+
+  /*
+  To check for suspected spam, we only care if _some_ of the regexes match our message, not all of them.
+
+  JavaScript has the `some` array method for just this purpose. `some` is a higher-order function that takes a callback as an argument. If the callback returns a truthy value for one or more elements of the array, `some` will return `true`. If no elements return a truthy value, `some` returns `false`.
+
+  For example:
+
+  ```javascript
+  const myArray = [2, 4, 6, 8, 99];
+
+  myArray.some(el => el % 2); // returns `true`, because one element is odd
+  ```
+
+  Uncomment the `return` statement and use `<array>.some` with `<regex>.test` to check if any of the `blacklistRegexes` match.
+  */
+};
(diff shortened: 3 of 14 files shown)


QuincyLarson:
@scissorsneedfoodtoo What do you think of this so far?

@lionel-rowe There probably doesn't need to be a UI element - it can just be a single JavaScript file that takes input and outputs whether it's spam. The list of concepts looks solid - we just want to make sure they spend ample time with each concept, and that there's a ton of repetition. We probably want this to ultimately be at least 100 tests long.

scissorsneedfoodtoo:
@lionel-rowe, sorry for the late response.

As @QuincyLarson mentioned, the concepts are solid. This is definitely off to a great start with a logical progression. We just need to break the steps down a bit more so there's more repetition. I'll leave some suggestions as review comments.

learn-regular-expressions-by-building-a-spam-filter/spamFilter040.js
```diff
@@ -0,0 +1,16 @@
+const isSpam = msg => {
+  const re = /viagra|free money|work from home|stock alert|dear friend/;
+  /*
+  Looking good!
+
+  However, the spammers are getting smarter. They've learned to sometimes mix up the case of their key phrases. As well as "free money", our filter also needs to be able to handle "FREE MONEY", "free MONEY", or even "fReE MoNEy".
```
scissorsneedfoodtoo: This can be simplified:

```
The spammers are getting smarter -- on top of "free money", our regex also needs to be able to handle "FREE MONEY", "free MONEY", or even "fReE MoNEy". Add the ignore case flag `i` to the end of `re` to deal with those variations. For example, to create a case-insensitive regex for "foo bar", you would simply write `/foo bar/i`.
```

learn-regular-expressions-by-building-a-spam-filter/spamFilter060.js
```diff
@@ -0,0 +1,27 @@
+const isSpam = msg => {
+  const blacklistRegexes = [
```
scissorsneedfoodtoo: This lesson could also be broken up. You could start by explaining that a single line of regex is not easy to read/maintain, and that an array to store everything would be better. Then users could build up the `blacklistRegexes` over several lessons before you explain a little about the `some` array. Your example code would be a good way to show `some` in action rather than write about the callback argument and truthy values.

learn-regular-expressions-by-building-a-spam-filter/spamFilter070.js
```diff
@@ -0,0 +1,22 @@
+const isSpam = msg => {
+  const blacklistRegexes = [
+    /viagra/i,
+    /free money/i,
+    /work from home/i,
+    /stock alert/i,
+    /dear friend/i
+  ];
+  /*
+  Great, refactoring is done, now back to business! Let's start handling those M4ngL3D stR1Ng5 ("mangled strings")!
```
scissorsneedfoodtoo: Try to simplify your instructions and break this down into multiple steps. It would be better to have the user change each key phrase in `blacklistRegexes` themselves. Introducing character classes and instructing them to change "i" in "viagra" and "dear friend" is a good first step.

learn-regular-expressions-by-building-a-spam-filter/spamFilter080.js
```diff
@@ -0,0 +1,29 @@
...
+    /v[i1|]agra/i,
+    /free money/i,
+    /work from home/i,
+    /stock alert/i,
+    /dear fr[i1|]end/i
+  ];
+  /*
+  Other characters can be mangled too. Luckily, we've managed to obtain a copy of _The Spammer's Handbook_ by S. Pam Merchant, which helpfully provides a lookup table of all the characters they use:
+
+  Character | Mangled versions
```
scissorsneedfoodtoo: The table here is really cool, but I think it would be better to have users create character classes one by one for each character. "i" is taken care of, so you could work through the rest of the list starting from "a".

learn-regular-expressions-by-building-a-spam-filter/spamFilter010.js
```diff
@@ -0,0 +1,16 @@
+const isSpam = msg => {
```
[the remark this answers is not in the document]
lionel-rowe: OK, I'll rework it to take a more gradually incremented approach.

I'm not sure about explaining arrow functions here, though - it doesn't have anything to do with regexes and would probably have already been explained by this point in the curriculum? If it hasn't, we can just use a function declaration - that way we won't end up re-explaining arrow functions in a load of unrelated challenges.

A variable declared with `const` but not assigned throws a SyntaxError. The closest approximation would be `const re = undefined`.

Also, the empty regex is `/(?:)/`, not `//` (the latter is a comment and will throw a SyntaxError if you try to assign it to variable). Perhaps would be better to start with this than `/./`, but then it requires explaining non-capturing groups. Or maybe simply `/a/` or some literal character is the simplest regex conceptually?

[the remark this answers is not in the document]
scissorsneedfoodtoo: Very good points here.

About the arrow functions, totally your call! Ideally users will work through the curriculum from beginning to end, but it's possible that they jump around. For cases like that it's good to review, at least briefly. I like the idea of using arrow functions here, and you don't have to go into them in depth. You could show an example and just leave it at that. Using a function declaration is fine, too.

I always forget about `const` throwing a SyntaxError if unassigned. Same with `//`. My mistake! Using `/a/` as a placeholder is a great idea. Mainly we want users to build as much of each project as possible. It's a bit too early in the project to explain non-capturing groups -- which probably aren't required at all -- so going with a single letter is probably best.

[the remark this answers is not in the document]
scissorsneedfoodtoo: On second thought, it would be better to start with `const re = /a/;`. It would give you some room to explain how the expression works at this point and would match the first "a" character in a string. Then you could expand on that by introducing the concept of matching whole words like viagra.

scissorsneedfoodtoo:
@lionel-rowe, would you mind if I merge this as is? After that, please feel free to continue breaking these lessons down and submit another PR when you can.

lionel-rowe:
@scissorsneedfoodtoo sure thing, go ahead. I'll try and get another PR up over the weekend.

scissorsneedfoodtoo:
@lionel-rowe, alright! Merging now. Looking forward to seeing the rest of your instructions.

PR #166, opened 2019-05-08
fix: add missing js project files

beaucarnes:
Some files were missing at the end of the JavaScript project. I'm adding them back.


scissorsneedfoodtoo:
Thank you for adding these in @beaucarnes, and @ryanwaltmurray again for your review. Just made a couple of minor changes including changing `var` to `let` in the code example in `index845`. Merging now :+1: :+1:

PR #168, opened 2019-05-13
Learn npm by Building an npm Module

dingjy94:


build-npm-module/index140.js
```diff
@@ -0,0 +1 @@
+// First create a `getUpperCase()` function which take a string variable `str` and return upper case of this string
```
beaucarnes: I think an example should be given so it is clear what format the function should be in. As you may know, when a user goes to the next step, the code shown will be the code we create in these challenges, not necessarily the code the user typed in. So we want to make sure the code the user types in is as close as possible to the 'correct' code.

Maybe something like:
// First create a `getUpperCase()` function which take a string variable `str` and return upper case of this string. For example, here is a `getTrimmed()` function which takes a string variable and returns the string with no whitespace on the end: `const getTrimmed = (str) => str.trim();`

build-npm-module/index170.js
```diff
@@ -0,0 +1,12 @@
...
+const getLowerCase = (str) => str.toLowerCase();
+
+const getSentenceCase = (str) => {
+  const lowerStr = getLowerCase(str);
+
+  return lowerStr.charAt(0).toUpperCase() + lowerStr.slice(1)
+};
+
+// Create getProperCase() function which take a string variable `str` and return proper case of this string.
+// This means you have to convert every word's first character to upper case and other characters to lower case.
```
beaucarnes: Consider breaking this up into a few steps.

build-npm-module/index190.js
```diff
@@ -0,0 +1,21 @@
...
+
+module.exports = {
+  getUpperCase,
+  getLowerCase,
+  getSentenceCase,
+  getProperCase
+};
+
+// Now you have finished your case-coverter module. Let's publish it to NPM.
+// You should first go to npmjs.com to get an account.
```
beaucarnes: We want this to work offline so we don't want to require external websites. Some things may need to be mocked. Maybe this could say something like:

Normally you would need to create your own account on npmjs.com. But for now, just use this demo account: username: camper, password: camper

dingjy94:
Hi @beaucarnes,
Thank you for your review. I had updated based on your suggestions. If there are any other changes need to make, just tell me.

scissorsneedfoodtoo:
Thank you for your review @beaucarnes, and @dingjy94 for making those changes so quickly! I'll take a quick look and leave any thoughts or suggestions as review comments.

build-npm-module/index270.js
```diff
@@ -0,0 +1,11 @@
+/**
+ * In your new directory's package.json file, there will be following line:
```
scissorsneedfoodtoo: This means you've installed your package successfully.

build-npm-module/index280.js
```diff
@@ -0,0 +1 @@
+// You should first import your package in main.js. Use node.js `require` to import a package.
```
scissorsneedfoodtoo: This could use an example of importing a module. I'm assuming this test description is appearing in the `main.js` file, and that the `case_converter` package is already installed:

Before you can use your package, you need to import it with Node's `require` keyword.
For example, `const arrLenthifier = require('arr_lengthifier');` imports the `arr_lengthifier` package and assigns it to a variable.
Import your `case_converter` package with `require` and assign it to the variable `caseConverter`.

build-npm-module/index290.js
```diff
@@ -0,0 +1,3 @@
+const caseConverter = require('./index');
+
+// Now let's try your functions. Create a varaible names `str` with a value `'hello free Code Camp!'`
```
scissorsneedfoodtoo: Now let's try your functions. Create a variable named `str` with a value `'hello free Code Camp!'`

build-npm-module/index300.js
```diff
@@ -0,0 +1,7 @@
+const caseConverter = require('./index');
+
+const str = 'hello free Code Camp!';
+
+// Let's try your getUpperCase function. Conver `str` to upper case and print it.
```
scissorsneedfoodtoo: This could use a quick example to show how to call the `getUpperCase` function:

Let's try your `getUpperCase()` function.
Here's an example that imports a module, calls a function, and prints the result to the console:
```
const arrLengthifier = require('arr_lengthifier');
const arr = [1, 2, 3];

console.log(arrLengthifier.getArrLength(arr));
```
Convert `str` to upper case and print it to the console.

build-npm-module/index290.js
```diff
@@ -0,0 +1,3 @@
+const caseConverter = require('./index');
```
scissorsneedfoodtoo: Just a note for myself. We'll eventually want to try and change this to `const caseConverter = require('case_converter');` and get this working locally somehow. Maybe a small npm module that's included in this directory?

build-npm-module/index310.js
```diff
@@ -0,0 +1,8 @@
+const caseConverter = require('./index');
+
+const str = 'hello free Code Camp!';
+console.log(caseConverter.getUpperCase(str));
+
+// Let's try your getLowerCase function. Conver `str` to lower case and print it.
```
scissorsneedfoodtoo: Let's try your `getLowerCase()` function. Convert `str` to lower case and print it to the console.

build-npm-module/index320.js
```diff
@@ -0,0 +1,9 @@
+const caseConverter = require('./index');
+
+const str = 'hello free Code Camp!';
+console.log(caseConverter.getUpperCase(str));
+console.log(caseConverter.getLowerCase(str));
+
+// Let's try your getProperCase function. Conver `str` to proper case and print it.
```
scissorsneedfoodtoo: Now let's try your `getProperCase()` function. Convert `str` to proper case and print it to the console.

build-npm-module/index330.js
```diff
@@ -0,0 +1,10 @@
+const caseConverter = require('./index');
+
+const str = 'hello free Code Camp!';
+console.log(caseConverter.getUpperCase(str));
+console.log(caseConverter.getLowerCase(str));
+console.log(caseConverter.getProperCase(str));
+
+// Let's try your getSentenceCase function. Conver `str` to sentence case and print it.
```
scissorsneedfoodtoo: Finally, let's try your `getSentenceCase()` function. Convert `str` to sentence case and print it to the console.

build-npm-module/index340.js
```diff
@@ -0,0 +1,9 @@
+const caseConverter = require('./index');
+
+const str = 'hello free Code Camp!';
+console.log(caseConverter.getUpperCase(str));
+console.log(caseConverter.getLowerCase(str));
+console.log(caseConverter.getProperCase(str));
+console.log(caseConverter.getSentenceCase(str));
+
+// You had learnt how to create, publish and use an npm module! CONGRATULATIONS!
```
scissorsneedfoodtoo: You've learned how to create, publish, and use an npm module! CONGRATULATIONS!

scissorsneedfoodtoo:
Thanks again for all of your hard work here @dingjy94! Will merge this now as is. Please submit another PR when you can to update the lessons mentioned above.

PR #181, opened 2019-05-28
change the server code entirely

padunk:


scissorsneedfoodtoo:
@padunk, thank you for your patience and your continued effort with this project! I'll take a look at all of your changes now and will leave suggestions as review comments if I notice anything.

node-js-web-server/server050.js
```diff
@@ -0,0 +1,13 @@
+const http = require('http');
+
+http.createServer(function (request, response) {
+   
+}).listen(8080);
+
+// now you can run your server by typing `node server` in your terminal
```
scissorsneedfoodtoo: We won't actually be able to run a server here, so we'll have to modify this text a bit:

While there's no way to run your server from this page, you could run it locally by saving it to a file called `server.js` and running `node server` from the same directory in your terminal.
However, if you did, there would be no way of knowing that your server is actually running.
Use `console.log()` to write your own notification that the server is running at port 8080.

node-js-web-server/server060.js
```diff
@@ -0,0 +1,10 @@
+const http = require('http');
+
+http.createServer(function (request, response) {
+   
+}).listen(8080);
+console.log('Server is running at http://localhost:8080');
+
+// now it time to check for request object.
```
scissorsneedfoodtoo: Inside the `createServer` callback function, write `console.log('request url: ', request.url);` to log incoming requests to the console.

node-js-web-server/server070.js
```diff
@@ -0,0 +1,30 @@
+const http = require('http');
+
+http.createServer(function (request, response) {
+   console.log('request url: ', request.url);
+}).listen(8080);
+console.log('Server is running at http://localhost:8080');
+
+// so we have this files structure:
```
scissorsneedfoodtoo: Our project has this file structure:
```
 ++public
    --index.html
    --products.html
    --about.html
    --404.html
    --forrest1.png
    --forrest2.png
    --forrest3.png
    --style.css
 --server.js
 --package.json
 --README.md
```
When users visit a certain URL, our server should respond with the correct page -- `http://localhost:8080/products` should get the response `./public/products.html`.
Create a variable called `filePath` and set it equal to `'./public' + request.url`.

scissorsneedfoodtoo: Just noticed that the code is using `var` rather than `let` or `const`. We could change this in the future.

node-js-web-server/server080.js
```diff
@@ -0,0 +1,14 @@
+const http = require('http');
+
+http.createServer(function (request, response) {
+   console.log('request url: ', request.url);
+   
+   var filePath = './public' + request.url;
+}).listen(8080);
+console.log('Server is running at http://localhost:8080');
+
+// now everytime user hit an url you could see the file path.
```
scissorsneedfoodtoo: Now when someone visits `http://localhost:8080/about`, `filePath` will be `./public/about`.
But what about the home page, `http://localhos:8080/`?
There's no file named `/`, so we will need to account for that.
Write an `if` statement to check if `filePath` is equal to `./public/`.
If it is, set `filePath` equal to `./public/index.html`.

node-js-web-server/server090.js
```diff
@@ -0,0 +1,15 @@
+const http = require('http');
+
+http.createServer(function (request, response) {
+   console.log('request url: ', request.url);
+   
+   var filePath = './public' + request.url;
+   if (filePath == './public/') {
```
scissorsneedfoodtoo: Just realized that this is using `==` rather than `===`. It's a minor detain, but we might consider changing this in the next draft.

node-js-web-server/server090.js
```diff
@@ -0,0 +1,15 @@
...
+      filePath = './public/index.html';
+   }
+}).listen(8080);
+console.log('Server is running at http://localhost:8080');
+
+// for this lesson require `path` module in your code
```
scissorsneedfoodtoo: We will need the `path` module to read the extensions of our files later.
Just under your `http` variable, `require` the `path` module and assign it to a variable named `path`.

node-js-web-server/server100.js
```diff
@@ -0,0 +1,17 @@
...
+   console.log('request url: ', request.url);
+   
+   var filePath = './public' + request.url;
+   if (filePath == './public/') {
+      filePath = './public/index.html';
+   }
+}).listen(8080);
+console.log('Server is running at http://localhost:8080');
+
+// we can get the extension name from filePath variable with
```
scissorsneedfoodtoo: We can get the extension name from the `filePath` variable with the `path.extname()` method.
Extract the extension name from `filePath` and store it in a variable named `extName`.

node-js-web-server/server110.js
```diff
@@ -0,0 +1,22 @@
...
+   var filePath = './public' + request.url;
+   if (filePath == './public/') {
+      filePath = './public/index.html';
+   }
+
+   var extName = path.extname(filePath);
+}).listen(8080);
+console.log('Server is running at http://localhost:8080');
+
+// you need to make sure that the value of `extName` is a string and lowercase.
```
scissorsneedfoodtoo: This lesson could probably be broken into two lessons, the first to make sure `extName` is lowercase and the second to ensure that it's a string.

scissorsneedfoodtoo:
Unfortunately I wasn't able to provide a detailed review of every step. I'll merge this for now and will leave more review comments over the next couple of days.

node-js-web-server/server220.js
```diff
@@ -0,0 +1,38 @@
...
+            
+        }); 
+      } else {
+         
+      }
+   });
+}).listen(8080);
+console.log('Server is running at http://localhost:8080');
+
+// now it is time for you to do something with the `response` object.
```
scissorsneedfoodtoo: Now it's time for you to do something with the `response` object.
To show the user if there's a 404 error, call the `writeHead` method from `response`.

node-js-web-server/server230.js
```diff
@@ -0,0 +1,40 @@
...
+            response.writeHead();
+        }); 
+      } else {
+         
+      }
+   });
+}).listen(8080);
+console.log('Server is running at http://localhost:8080');
+
+// `writeHead` method can be passed two arguments:
```
scissorsneedfoodtoo: The `writeHead` method takes two arguments: the status code of the response, and a headers object.
Pass `writeHead` 404 as its first argument, and `{'Content-Type': // file content-type}` as the second argument.
Remember that you can get the content type for the second argument from the `contentType` variable.

node-js-web-server/server240.js
```diff
@@ -0,0 +1,36 @@
...
+            response.writeHead(404, { 'Content-Type': contentType });
+        }); 
+      } else {
+         
+      }
+   });
+}).listen(8080);
+console.log('Server is running at http://localhost:8080');
+
+// you also need to tell the server that the response is ended.
```
scissorsneedfoodtoo: Then you need to tell the server that the response has ended.
Just after the `writeHead` method, call the `end` method from `response`.

node-js-web-server/server250.js
```diff
@@ -0,0 +1,40 @@
...
+            response.end();
+        }); 
+      } else {
+         
+      }
+   });
+}).listen(8080);
+console.log('Server is running at http://localhost:8080');
+
+// this `response.end()` method MUST be called after each response.
```
scissorsneedfoodtoo: `response.end()` **must** be called after each response.
Pass `end` two arguments: `content`, and `'utf-8'`.
Remember that `content` is the data from the `404.html` file being read, and `'utf-8'` is the encoding of the file.

node-js-web-server/server260.js
```diff
@@ -0,0 +1,41 @@
...
+            response.end(content, 'utf-8');
+        }); 
+      } else {
+         
+      }
+   });
+}).listen(8080);
+console.log('Server is running at http://localhost:8080');
+
+// for this last challenge, write the else statement body.
```
scissorsneedfoodtoo: Finally, write the `else` block to finish the server.
Use the `writeHead` and `end` methods just like you did in the `if` block.
However, since this response indicates a successful request to the server, pass the OK status code 200 as the first argument of `writeHead`.

node-js-web-server/server070.js
padunk: > Just noticed that the code is using `var` rather than `let` or `const`. We could change this in the future.

I didn't notice it. I only change the module with `const`. I think we can change it right away.

scissorsneedfoodtoo: Okay! Changing that along with the edited test descriptions I suggested would be really great. Please just submit another PR whenever you can :+1: :+1:

PR #186, opened 2019-06-24
update var to const and instructions

padunk:
updating:
1. `var` to `const`
2. `==` to `===`
3. instructions


scissorsneedfoodtoo:
@padunk, thank you for your patience. I looked over your new instructions and made a few changes to teh spacing. I also split up a few of the lessons into several steps. Everything LGTM!

Issue #104, opened 2019-03-08
Learn RESTful APIs by Building a Grocery List App


QuincyLarson:

picolloo:
I'm interessed to participate in this one

QuincyLarson:
@picolloo OK - cool. You have expressed interest in several of these. I recommend tackling these one at a time. Which project do you want to build first?

picolloo:
So I'll finisih those I'm involved with and help with this one later if it's necessary

ritute:
Is the type of service set in stone for this, or can we suggest a different one (such as a recipe book) that would be simpler and translate better to a typical application, since weather is something usually read from another service (openweathermap) and would be more unusual to implement (it also consists largely of numeric data that would make the API more cumbersome in testing and potentially detract from the goal of this learning project).

QuincyLarson:
@ritute the service isn't set in stone. And we won't be using a real API - we'll just use dummy data for reliability, and so that the curriculum can still be completed without internet access (we are gradually working toward making freeCodeCamp an offline-first application.)

scissorsneedfoodtoo:
@ritute, just wanted to check in and see if you've made any progress on this. Please let me know if there's anything I can help with.

shimphillip:
@scissorsneedfoodtoo 

I am also interested in taking this one. Can I take two projects at a time?

scissorsneedfoodtoo:
@shimphillip, sorry, but it's best if you just take one at a time. Which project would you like to work on first? This one, or the shopping cart?

beaucarnes:
@shimphillip Are you going to work on this one? If not, I can give it a try.

shimphillip:
@beaucarnes Sure you can take it.

beaucarnes:
I've been working on this a bit and have been thinking about the best way to show aspects of a RESTful API. I think @ritute was right with the comment about how a weather API is not the best project for this. Weather APIs mainly only accept GET requests. But to really learn about RESTful APIs people should see all the basic CRUD routes. 

In my opinion, the MongoDB project (issue #103 ) is already the exact thing that this project should be. That project is a RESTful API. I think we should consider removing this project from the curriculum.

scissorsneedfoodtoo:
[...]

PR #187, opened 2019-06-27
Restful API project

beaucarnes:
Project and tests for Learn RESTful APIs by Building a Grocery List App #104


scissorsneedfoodtoo:
@beaucarnes, awesome! Thank you for taking the time to write these up. The only things I noticed is that it seems like the instructions repeat in `server070.js`, and there might be a step or two missing for the last line on `server075.js`, `app.listen(port, () => console.log("List server listening on port " + port));`. If so, they can be addressed later. Thanks again :+1: :+1:

PR #188, opened 2019-06-29
[WIP] feat: first draft of building view counter

### LINKED ISSUE #111: Learn AJAX by Building a View Counter

karuppiah7890:
Closes #111 

This is Work In Progress PR, to get reviews on code continuously instead of the last minute

### DIFF
--- a/building-view-counter/index030.html
+++ b/building-view-counter/index030.html
@@ -1,0 +1,37 @@
+<!DOCTYPE html>
+<!--
+    Now we can see the text in our web page but we haven't populated the count yet.
+
+    Now to get the count, we will be sending an ajax request to the server in our
+    javascript code and it will take a moment to receive the response. Till we get
+    the response, we will not able to show the count, so, in this moment
+    we can hide the text when the web page loads, and when we get the view
+    count from server we can show the text along with the count
+
+    To hide the text when the web page loads, wrap it around a <span> tag,
+    give it an id "message". Use css property "display" to hide the <span>
+ -->
+<html lang="en">
+<head>
+    <title>What's Ajax?</title>
+    <link rel="stylesheet" href="css/style.css">
+</head>
+<body>
+    <h1>What's Ajax?</h1>
+    <p class="article">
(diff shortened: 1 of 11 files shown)


building-view-counter/index030.html
```diff
@@ -0,0 +1,37 @@
+<!DOCTYPE html>
+<!--
+    Now we can see the text in our web page but we haven't populated the count yet.
```
scissorsneedfoodtoo: This could be broken up into two steps. The first could instruct the learner to wrap the text in a span element, and the second could be adding the rule to the CSS file.

The first step could read:

Now we can see the text, but we haven't populated the count yet.
To get the count, we will send an AJAX request to the server, but it will take a moment to receive a response.
We should hide the text until we get the count from the server.
Wrap the text you just added in a `<span>` element and give it the `id` "message".

The next step should appear in the CSS file and could read:

Create a new rule targeting the "message" `id` and set `display` to `none`.

building-view-counter/index040.html
```diff
@@ -0,0 +1,37 @@
+<!DOCTYPE html>
+<!--
+    Now we see that the span tag is not displayed.
```
scissorsneedfoodtoo: Now we see that the `<span>` element is not displayed.
To display the view count with the current code, we have to get all of the text in the `<span>` and append the count to the end.
It would be easier if we can populate a field for the count right away.
To do this, add another `<span>` element at the end of the one you just added, and give it the `id` "viewCount".

building-view-counter/index050.html
```diff
@@ -0,0 +1,29 @@
+<!DOCTYPE html>
+<!--
+    Now we are all good with html code, we can work on the javascript code to
```
scissorsneedfoodtoo: This step could be removed. Instead, learners could see an empty JS file with commented out instructions.

scissorsneedfoodtoo:
@karuppiah7890, things are really coming along here. I know this is still a draft, but I thought I would give some early feedback. Looking forward to seeing the rest of your steps :+1: :+1:

PR #189, opened 2019-07-05
Shopping Cart update

shimphillip:


scissorsneedfoodtoo:
Thanks for all of your hard work with these edits, @shimphillip. Hope you don't mind, but I went in and moved some of your early test descriptions around so they're near the code that will change. I also did things like add backticks for code blocks and lines of code.

learn-data-structures-by-building-a-shopping-cart/index16a.js
```diff
@@ -0,0 +1,61 @@
...
+    this.items.splice(index, 1);
+  }
+  
+  clearCart() {
+    this.items = [];
+  }
+
+  applyDiscount(amount) {
+    /*
+    Inside the applyDiscount function, calculate the discount amount using discount percentage we previously declared. For ex:
```
scissorsneedfoodtoo: Inside the `applyDiscount` function, calculate and return the discount by using the `discountPercentage` we previously declared. 
You can use the formula `discountPercentage / 100 * amount`.
Don't forget to use the `return` and `this` keywords!

learn-data-structures-by-building-a-shopping-cart/index16b.js
```diff
@@ -0,0 +1,65 @@
...
+    this.items = [];
+  }
+  
+  applyDiscount(amount) {
+    return this.discountPercentage / 100 * amount
+  }
+}
+
+/*
+Let's print to see the output of applyDiscount method!
```
scissorsneedfoodtoo: Let's print to the console to see the output of the `applyDiscount` method.
```
  const shoppingCart = new ShoppingCart();
  console.log(shoppingCart.applyDiscount(2.99));
```

learn-data-structures-by-building-a-shopping-cart/index17.js
```diff
@@ -50,20 +50,17 @@ class ShoppingCart {
...
   
   applyDiscount(amount) {
-    return this.discountPercentage / 100 * amount
-  }
-}
-
-/*
-Notice that the returned amount of appliyDiscount function has a lot of numbers after the decimal.
+    /*
+    Notice that the discount money has more than 2 numbers after the decimal point.
```
scissorsneedfoodtoo: Notice that the expected discount has more than 2 numbers after the decimal point.
    This is common to see when working with numbers in JavaScript.
    There is in-built method called `toFixed()` that can help us.
    This method takes a number as a parameter for the amount of digits that you want to appear after the decimal point.
    It will also round up the number if necessary.
    Wrap the formula in parentheses and chain `toFixed(2)` to the end of it.

learn-data-structures-by-building-a-shopping-cart/index17a.js
```diff
@@ -0,0 +1,65 @@
...
+    this.items = [];
+  }
+  
+  applyDiscount(amount) {
+    return (this.discountPercentage / 100 * amount).toFixed(2);
+  }
+}
+
+/*
+toFixed method inadvertently converts a number into a string.
```
scissorsneedfoodtoo: One thing to keep in mind is that the `toFixed()` method converts numbers into strings.
Let's log a discounted price to the console and check the output:
```
const shoppingCart = new ShoppingCart();
console.log(shoppingCart.applyDiscount(2.99));
console.log(typeof shoppingCart.applyDiscount(2.99));
```

learn-data-structures-by-building-a-shopping-cart/index17b.js
```diff
@@ -0,0 +1,67 @@
...
+    this.items = [];
+  }
+  
+  applyDiscount(amount) {
+    return (this.discountPercentage / 100 * amount).toFixed(2);
+  }
+}
+
+/*
+You will now see that the number has been rounded and its type is string!
```
scissorsneedfoodtoo: As you can see, the number has been rounded to two decimal places and its type is `"string"`!
We still want to work with numbers, so use the built-in `parseFloat()` method to convert the string back to a number.
For example: 
```
console.log(parseFloat((1 / 3).toFixed(2))); // 0.33
```

scissorsneedfoodtoo:
Just left a few more suggestions. Thanks again @shimphillip. Looking forward to seeing the rest of your steps :+1: :+1:

PR #190, opened 2019-07-09
second draft for learn npm module

dingjy94:


build-npm-module/index175.js
```diff
@@ -8,7 +8,6 @@ const getSentenceCase = (str) => {
   return lowerStr.charAt(0).toUpperCase() + lowerStr.slice(1)
 };
 
-const properCaseArr = (str) => str.split(' ').map((word) => getSentenceCase(word));
-
-// Create getProperCase() function to join properCaseArr into spce separated sentence.
+const getProperCase = (str) => {};
 
+// First we can split the input string by space, and call our `getSentenceCase` on each word, store the result array in a local variable called `properCaseArr`.
```
scissorsneedfoodtoo: This one might also benefit from being broken down into two steps. Still, here's the current text:

Split the input `str` into an array of words, call `getSetenceCase` on each word, and store the resulting array as a local variable called `properCaseArr`.

build-npm-module/index180.js
```diff
@@ -8,7 +8,9 @@ const getSentenceCase = (str) => {
   return lowerStr.charAt(0).toUpperCase() + lowerStr.slice(1)
 };
 
-const properCaseArr = (str) => str.split(' ').map((word) => getSentenceCase(word));
-const getProperCase = (str) => properCaseArr.join(' ');
+const getProperCase = (str) => {
+  const properCaseArr = (str) => str.split(' ').map((word) => getSentenceCase(word));
+};
+
+// Call join on `properCaseArr` to convert the array to a string and return the result.
```
scissorsneedfoodtoo: Call the `join()` method on `properCaseArr` to convert the array into a string and return the result.

build-npm-module/index185.js
```diff
@@ -0,0 +1,17 @@
...
+  return lowerStr.charAt(0).toUpperCase() + lowerStr.slice(1)
+};
+
+const getProperCase = (str) => {
+  const properCaseArr = (str) => str.split(' ').map((word) => getSentenceCase(word));
+
+  return properCaseArr.join(' ');
+};
+
+// You should then use node.js `module.exports` to export these four functions so that users can import and use your package's function
```
scissorsneedfoodtoo: This may be the first time learners use `module.exports`, so we should show an example:

Use `module.exports` to export your four functions so that users can import and use them after installing your npm package.
For example:
```
module.exports = {
  getArrLength,
  sumArr
};
```

build-npm-module/index190.js
```diff
@@ -17,6 +17,8 @@ module.exports = {
   getProperCase
 };
 
-// Now you have finished your case-coverter module. Let's publish it to NPM.
-// Normally you would need to create your own account on npmjs.com.
-// But for now, just use this demo account: username: camper, password: camper
+/*
+Now that you have finished your case_converter module, let's publish it to the npm registry.
```
scissorsneedfoodtoo: This test description should come later when learners have to enter `npm adduser`. It might be better to have students edit the `README.md` file, then login before publishing the package. Either way, here's the edited text for this step:


Now that you have finished your `case_converter` module, let's publish it to the npm registry.
Normally you would need to create your own account on https://www.npmjs.com.
But for now, just use this demo account: 
```
username: camper
password: camper
```

build-npm-module/index200.js
```diff
@@ -1,5 +1,7 @@
-/**
- * Create a README.md under your case_converter directory.
- *
- * This will be the description and instruction file for your package. It will shou on your module's npm home page.
- */
+/*
+Let's edit your README.md under your case_converter directory.
```
scissorsneedfoodtoo: For this step we can imagine that the `README.md` file is already created and open. We could break the next step down further by prompting the learner to enter the markdown text themselves.

scissorsneedfoodtoo:
Thanks again for all your hard work! This draft looks great. A few lessons could be broken up a bit more, but other than that, everything LGTM. Also, your idea during our call to better introduce commands like `cd` and `mkdir` was really good. Looking forward to seeing that in the next draft :+1: :+1:

Issue #118, opened 2019-03-08
Learn Functional Programming by Building a Spreadsheet


QuincyLarson:

codingk8:
OK to contribute to this one.

QuincyLarson:
@codingk8 OK - cool. See if you can build a very basic spreadsheet in JavaScript on a site like CodePen and we can go from there. This is going to be a particularly exciting project.

codingk8:
I'm currently busy with a CSS project. Feel free to come and move on with this one in the meantime.

QuincyLarson:
@codingk8 OK - thanks for confirming. This one project is open for contributors.

RandellDawson:
@QuincyLarson I will take this one instead of the CVS one if that is ok.

QuincyLarson:
@RandellDawson Sure thing - I've assigned it to you. Best of luck!

scissorsneedfoodtoo:
@RandellDawson, just checking in to see how things are going with your prototype. Do you have anything you can show us? If so, we could give some early feedback.

RandellDawson:
@scissorsneedfoodtoo I have been busy with my job search, so I have not spent as much time on it as I had hoped.  If you want someone else to take on this project who can spend, I am fine with that.

scissorsneedfoodtoo:
@RandellDawson, no problem! Thank you for letting us know. I'll put the help wanted label back on for now, but please feel free to pick it up again later if it's still unclaimed and you have some time. Good luck with your job search!

UsamaHameed:
Hello, I am back from vacation and can see that the typescript project is in good hands. I think I can take on this one instead. What kind of project do you want here? How versatile should it be? Can we limit it to some simple operations?

scissorsneedfoodtoo:
@UsamaHameed, sure, the project is all yours! As @QuincyLarson mentioned above, a very simple spreadsheet on a platform like CodePen is totally fine. My thinking is that the spreadsheet doesn't have to be too versatile, and limiting it to a few simple operations is fine. The main goal here is to teach functional programming, so we're good as long as we can do that.

scissorsneedfoodtoo:
@UsamaHameed, just wanted to check in and see how everything's going with this project. Were you able to make some progress on a prototype?

sw-yx:
as someone who's tried doing this.. we may need to limit the scope a bit. spreadsheets are very hard to make right.

UsamaHameed:
[...]

PR #218, opened 2019-09-24
feat: third draft of spreadsheet (adds tests)

SomeDer:
issue #118


RandellDawson:
I suggest using a helper function in the After Test or Before Test section which strips all the white space from the code to simplify your regular expressions in the tests.  Having all of those `\s*` can be avoided doing this.

SomeDer:
@RandellDawson I've actually done this for all tests after the first few, but I've now done this for those tests as well.

scissorsneedfoodtoo:
Thank you for all of your hard work here, @SomeDer. Really looking forward to going through all of your steps.

Just wondering, did you create a `meta.json` file to get this working locally? If so, would it be possible to submit these files as a PR to the main fCC repo? The current plan is to merge them into the master branch with an `isBeta` flag so they're only accessible in a dev environment. If not, let's merge them here and I can move everything over to the main repo.

PR #223, opened 2019-10-02
Chat Websocket - step by step

picolloo:


scissorsneedfoodtoo:
@picolloo, thank you again for you PR. Just to give you an update, I'm slowly going through this and will leave some detailed review comments before merging.

steps/server02.js
```diff
@@ -0,0 +1,5 @@
+// Create a PORT variable and assign it to 3000, this will be the port your server will listen to
```
scissorsneedfoodtoo:
```
// Create a `PORT` variable and set it equal to 3000. This will be the port your server listens to.
// Call the app's `listen` method by passing it `PORT` as an argument.
```

steps/server03.js
```diff
@@ -0,0 +1,10 @@
+// Insert a callback function as the second parameter of listen that prints "Server bound to the port <PORT>" on the console
```
scissorsneedfoodtoo:
```js
// Add a callback function as the second parameter of `listen` that prints "Server bound to port <PORT>" to the console.
// After that, you should be able to run the command `node server.js` and see the message you just wrote.
```

steps/server10.js
```diff
@@ -0,0 +1,28 @@
+// Now the socket io is applied to our server, everything is configured
```
scissorsneedfoodtoo: I think there might be a step missing here, but no matter. `io.on("connection"...` could be broken up into two small steps to break up this good explanation:

```
// Every Socket.IO interaction begins with the socket connection from the client to the server.
// To listen to the connection event we have to call the `on` method from the `io` instance we created.
// Chain `on` to `io` and pass it "connection" to listen for any incoming connection events.
```

The next step could read:
```
// Pass a callback function as the second parameter to `on` and have it log "User connected" to the console whenever the connection event is fired.
```

steps/server11.js
```diff
@@ -0,0 +1,27 @@
+// After open the localhost:3000 in the browser
```
scissorsneedfoodtoo: It might not be possible to see the server-side logs when this is in production. Either way, I tried to simplify this a bit:

```
// The callback you wrote actually receives one argument -- the socket that triggered the event.
// Pass `socket` as a parameter to the callback function you just wrote and log `Socket connected: <id of the socket>` to the console instead of "User connected".
```

websocket-chat/public/script.js
```diff
@@ -0,0 +1,75 @@
+const socket = io("http://localhost:3000");
```
scissorsneedfoodtoo: In the final version of this project learners should also write all of this code. They can start by setting up a basic server then work on this file to get the different functions working.

steps/server12.js
```diff
@@ -0,0 +1,37 @@
+// Now we have to create the methods for the front to communicate with the backend
```
scissorsneedfoodtoo: Learners should change something each step. As I wrote in another comment, in the final version we should have learners write as much of the code as possible, even on the front end. I imagine that when learners get to this point they could start working on `/public/script.js` to get things working between the front and backend.

steps/server13.js
```diff
@@ -0,0 +1,28 @@
+// You should try to test if sendMessage function is working
```
scissorsneedfoodtoo: Once learners worked on `/public/script.js` to the point where the button should work, they can come back here to handle the `"newMessage"` event here. When they do, we could split this up into two or three shorter steps.

steps/server14.js
```diff
@@ -0,0 +1,35 @@
+// Now you should be able to see the message you wrote on the chat printed out in the server console
```
scissorsneedfoodtoo: The information here is good, but it could be broken up into two shorter steps so we have more time to properly explain how things should work.

scissorsneedfoodtoo:
@picolloo, thank you for your patience.

Everything is looking good on the server side. I just left some general comments that we can work into the next draft. Generally we should try to split up lessons a bit more. Though learners by this point should have a good grasp of Node and Express, it would give us more room to explain how and why we're doing certain things.

Also, your code on the frontend side of things looks great. Could you start breaking that up into steps, too? I think they could be inserted between already existing steps, so learners could work through up to `server11.js`, then go to `script12.js` and start building up the code there. Then just the last couple of `server__.js` files can be renumbered.

PR #224, opened 2019-10-07
feat: steps for the blog file

RafaelDavisH:
Started with the first steps for the blog page HTML file.

### DIFF
--- a/learn-CSS-by-Building-a-blog/blog39.html
+++ b/learn-CSS-by-Building-a-blog/blog39.html
@@ -1,0 +1,4 @@
+<!-- 
+  We are going to use the same structure as the previous HTML page.
+  Copy and paste here ALL the content inside the `index38.html` file.
+ -->
--- a/learn-CSS-by-Building-a-blog/blog40.html
+++ b/learn-CSS-by-Building-a-blog/blog40.html
@@ -2,7 +2,7 @@
 <html>
 <head>
     <meta charset="UTF-8">
-    <title>Home | Learn Basic CSS by Building a Blog</title>
+    <title>Home | Learn Basic CSS by building a Blog</title>
     <link rel="stylesheet" href="css/style.css">
     <link href="https://fonts.googleapis.com/css?family=Playfair+Display:700|Source+Sans+Pro:300,400&display=swap" rel="stylesheet">
 </head>
@@ -39,8 +39,6 @@
                         metus. Suspendisse potenti. Aenean at varius ex. Interdum et malesuada 
                         fames ac ante ipsum primis in faucibus. Aliquam ut nisl metus. Nam 
                         bibendum turpis et vestibulum semper
-
-                        <!-- Next step -->
                     </p>
                     <p>
                         Lorem ipsum dolor sit amet, consectetur adipiscing elit. Sed imperdiet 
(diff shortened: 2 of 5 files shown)


learn-CSS-by-Building-a-blog/blog39.html
```diff
@@ -0,0 +1,4 @@
+<!-- 
+  We are going to use the same structure as the previous HTML page.
```
scissorsneedfoodtoo: In the final version of the project we won't be using the individual file names, just `index.html`, and perhaps `article1.html` and `article2.html`. If this is for `article1.html`, this step could read:

```
<!--
  We are going to use the same structure as the previous HTML page.
  Copy all the content inside the `index.html` file and paste it here.
-->
```

learn-CSS-by-Building-a-blog/blog40.html
```diff
@@ -75,4 +73,14 @@
         </section>
     </div>
 </body>
-</html>
\ No newline at end of file
+</html>
+
+<!--
```
scissorsneedfoodtoo:
```
<!-- 
  Delete all the content nested inside the `section` element with the class `container`.
  Here is an example of how it should look:

  <section class="main">
    
  </section>
 -->
```

learn-CSS-by-Building-a-blog/blog42.html
```diff
@@ -0,0 +1,31 @@
...
+            </nav>
+        </header>
+        <section class="container">
+            <div></div>
+        </section>
+    </div>
+</body>
+</html>
+
+<!--
```
scissorsneedfoodtoo:
```
<!-- 
  Nest a `div` element inside the `section` and give it the class `blog`.
-->
```

learn-CSS-by-Building-a-blog/blog42.html
```diff
@@ -0,0 +1,31 @@
...
+                    <li><a href="#">Intro</a></li>
+                    <li><a href="#">Blog</a></li>
+                </ul>
+                <a href="#" id="logo" >
+                    <img src="https://d33wubrfki0l68.cloudfront.net/e66392bc2941ad564b6956321e778f29073cf585/4190b/img/glyph-alternative.svg" alt="freeCodeCamp logo">
+                </a>
+            </nav>
+        </header>
+        <section class="container">
+            <div></div>
```
scissorsneedfoodtoo: The `div` appears here suddenly. It can be removed so learners can add it and give it a class themselves.

scissorsneedfoodtoo:
@RafaelDavisH, thank you for your patience. All these steps look good! Looking forward to seeing the rest of your steps breaking down both articles and adding the CSS.

learn-CSS-by-Building-a-blog/blog39.html
RafaelDavisH: Le all the HTML files are going to be together. Wouldn’t this make it confusing to distinguish between the actual index.html and the “blog.html”? Should we have a folder for all index.html, “blog.html” (index.html), and so on?

scissorsneedfoodtoo: @RafaelDavisH, there's still some things we need to figure out about how the new platform will work, particularly with multiple files. The idea now is that we'll have a way to iterate through the different versions of each file in the background (`index01.html`, `index02.html`), but the learners will only see `index.html`.

I think the file structure is fine for now, but we can always change it in the future. In the test descriptions you're writing now we can just refer to each file by the name learners will see like `index.html`.

RafaelDavisH: got it 👍

Issue #108, opened 2019-03-08
Learn D3 by Building a Map of Earth


QuincyLarson:

QuincyLarson:
@HKuz This is a tentative project idea - really anything that involves D3 would be a good project.

scissorsneedfoodtoo:
@HKuz, just wanted to check in and see how things are going with this project. Were you able to start working on a prototype?

HKuz:
Hi @scissorsneedfoodtoo - just started digging into this one now that JSON is in good shape! I'm reviewing some D3 now (little rusty 😅 ) and creating an outline, and hope to start a skeleton project by the weekend 👍

scissorsneedfoodtoo:
@HKuz, sounds good! Really looking forward to seeing your project. It'll be good review for me since I forgot almost everything about D3!

HKuz:
Alright, here's a skeleton of this project [here](https://hkuz.github.io/fCC_D3_CurrExp/), with code repo [here](https://github.com/HKuz/fCC_D3_CurrExp). I would have liked to have gotten a little farther along, but D3v5 threw me and the example code I was using a few curveballs 😅 .

This is a start / WIP, with the following things I want to add in:
- Find country data (like population) to also bind to the map
- Add tooltip functionality to display the country name and [population] data
- Color the countries based on that data to teach scales
- Add pan/zoom functionality (not critical, but seems like a common function in interactive maps)

scissorsneedfoodtoo:
@HKuz, looks great so far! Showed your prototype to @QuincyLarson and we're both looking forward to seeing what you do with this. He suggested that you include the world map JSON, along with whatever other data sets you'll use, in your project directory. We should be able to host everything on the platform, and it will be nice for students to open up those files and take a look at them.

Your ideas all sound good, and it would be great if you could find a way to include a bar graph somehow since that's a common use for D3.

HKuz:
[...]

PR #277, opened 2019-12-19
feat: add D3 Map of World project lessons, supporting files

HKuz:
Hi all,

Here's my holiday PR for the D3 Map of the World project lessons, along with data and supporting files. I did my best to keep lesson text concise, but some concepts like geospatial data formats, projections, or D3's enter-update-exit methods needed a certain level of explanation. I'm certainly open to feedback here. Also, given the number of lessons (130 total) and the fact most of the coding is done in the `.js` files, I didn't copy the `.html` and `.css` files for every lesson - the last one can be carried forward until a lesson touches it again if that's what the platform requires.

Live project is still up on a personal repo [here](https://hkuz.github.io/fCC_D3_CurrExp/).

Addresses Issue #108 

Happy holidays! 🎁 🎄 🌟


scissorsneedfoodtoo:
@HKuz, thanks for your patience and hope you enjoyed the holidays.

I think you're right, it will be very difficult to explain certain concepts like the ones you listed in just a few sentences per step. 

Let's merge this to get all of your hard work in, and I'll go back and leave suggestions as review comments here for a next draft.

d3-map-of-world/Lessons/0020_index.html
```diff
@@ -0,0 +1,24 @@
...
+    <title>World Population</title>
+  </head>
+  <body>
+    <!--
+    D3 is a library, which means it packages its visualization functionality into
+    a script that you can download and save in your project folder, or link to
+    with a CDN. Before you can use D3's methods in your own script, you need to
+    tell the browser where this script is.
+
+    We saved a copy of D3's version 5 code in the project folder for you, in a
```
scissorsneedfoodtoo: To keep things as concise as possible, how about the following?:

```
We saved a copy of D3's version 5 code in the project folder for you, in a file named `d3v5min.js`. Before you can use D3's methods in your own script, you need to tell the browser where this script is.

Create a `script` tag with the `type` attribute set to `text/javascript` and the `src` attribute set to the name of the D3 script file.
```

d3-map-of-world/Lessons/0030_index.html
```diff
@@ -0,0 +1,30 @@
...
+    <meta http-equiv="Content-Type" content="text/html; charset=utf-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0, maximum-scale=1.0, user-scalable=no">
+    <title>World Population</title>
+  </head>
+  <body>
+    <!-- D3 v5 script -->
+    <script type="text/javascript" src="./d3v5min.js"></script>
+
+    <!--
+    There are many different file formats that hold geospatial data that you
```
scissorsneedfoodtoo: Really great information here. How about the following to shorten this description a bit?:

```
There are many different file formats that hold geospatial data that you can use to create maps. We're only going to focus on two that are web-friendly: GeoJSON and TopoJSON. For now, understand that they hold information that describes the shape of regions, states, provinces, and countries, among other things.

D3 provides a script that gets the necessary information from those formats so you can create a map. A copy of this script is saved in the project folder and named `topojsonv1min.js`.
    
Create another `script` tag with the `type` attribute set to `text/javascript` and the `src` attribute set to the name of the TopoJSON script file.
```

d3-map-of-world/Lessons/0070_index.js
```diff
@@ -0,0 +1,14 @@
+// Setup
+
+
+/*
+As noted in the first lesson, D3 is designed to work with common web standards,
```
scissorsneedfoodtoo: Since the note about web standards will be moved to the intro page, perhaps this could read:

```
This project will use D3 to draw a map with SVG, which you'll add to your page with this script. SVG elements have `height` and `width` attributes, so the browser knows the size and how to display them.

Using the `const` keyword, declare an `svgWidth` variable and set it to `1000`, and declare an `svgHeight` variable and set it to `700`.
```

d3-map-of-world/Lessons/0080_index.js
```diff
@@ -0,0 +1,23 @@
+// Setup
+const svgWidth = 1000;
+const svgHeight = 700;
+
+
+/*
+Since you linked to a local copy of the D3 script in the `HTML` file, all of D3's
```
scissorsneedfoodtoo:
```
D3 has a `.select()` method that allows you to grab an HTML element in the DOM and work with it. For example, you would select the first item with the class `hello` with `d3.select(".hello")`.

You'll need the method here to add an `svg` element to the page to hold your map.

Using the `const` keyword, declare a variable named `svg`. Set it equal to the D3 `select` method, and pass the method the `map` `id` of the `div` you created.
```

d3-map-of-world/Lessons/0100_index.js
```diff
@@ -0,0 +1,23 @@
+// Setup
+const svgWidth = 1000;
+const svgHeight = 700;
+
+const svg = d3.select("#map")
+    .append("svg")
+
+
+/*
+The last piece to create the `svg` element is to tie `width` and `height`
```
scissorsneedfoodtoo:
```
The last step to create an `svg` element is to assign `width` and `height` attributes to it. D3 has an `.attr()` method for this, and it takes two arguments. The first argument is the name of the attribute as a string and the second argument is the value for it.

Chain two `attr` methods right after the `append` method. The first should add a `"width"` attribute with the value set to your `svgWidth` variable. The second should add a `"height"` attribute with the value set to your `svgHeight` variable.

Since there isn't a unit specified with these values, D3 will interpret them as pixels by default.
```

d3-map-of-world/Lessons/0280_index.js
```diff
@@ -0,0 +1,39 @@
...
+const svgHeight = 700;
+
+const svg = d3.select("#map")
+    .append("svg")
+    .attr("width", svgWidth)
+    .attr("height", svgHeight);
+
+
+
+/*
```
scissorsneedfoodtoo: Did my best to shorten this description. Let me know if it still makes sense:

```
The coordinates in the TopoJSON features describe countries as they are on Earth's three-dimensional surface. You need to translate them so the countries display properly on your two-dimensional `svg`. A map **projection** is a method for making this conversion.

Projections have to make adjustments to convert 3D coordinates onto a 2D surface, so they all have some form of distortion. They generally try to preserve one aspect - some keep areas accurate, some distance, and some the angles between places (good for nautical navigation).

For this map, you'll use a classic Mercator projection. Using the `const` keyword, declare a variable named `projection` and set it to `d3.geoMercator()`.
```

d3-map-of-world/Lessons/0290_index.js
```diff
@@ -0,0 +1,38 @@
...
+const popPath = "./worldPopulation.csv";
+
+// Setup
+const svgWidth = 1000;
+const svgHeight = 700;
+
+
+
+/*
+D3 provides a useful `.scale()` method that you can chain to the Mercator
```
scissorsneedfoodtoo:
```
D3 provides a useful `.scale()` method that you can chain to the Mercator projection, which scales the map larger or smaller when it first loads. You pass a number into the method, and this "scale factor" corresponds to the distance between projected points. In other words, the larger the value, the more "zoomed in" the map appears.

The default value for the Mercator projection is 150, which leaves the map "cut off' with extra ocean on the left and right. Instead, you're going to use a formula based on the `svgWidth` to fill the entire width of your `svg`.

Using the `const` keyword, declare a variable named `scale` and set it to the `svgWidth` divided by `(2 * Math.PI)`.
```

d3-map-of-world/Lessons/0320_index.js
```diff
@@ -0,0 +1,37 @@
...
+    .attr("height", svgHeight);
+
+const projection = d3.geoMercator()
+    .scale(scale)
+    .translate([svgWidth / 2, svgHeight / 2])
+
+
+
+/*
+The last part of the `projection` declaration will use D3's `.center()` method.
```
scissorsneedfoodtoo:
```
The last part of the `projection` declaration will use D3's `.center()` method. It takes an array of `[longitude, latitude]` coordinates, and centers the projection there.

D3 will use the default value of `[0°, 0°]`, so this method allows you to put any country or place in the center.

For a Mercator projection, 0° longitude works great, but we won't need to see as much of Antarctica as Greenland, so we'll move the center up to 20° latitude.

Chain the `.center()` method to your projection and pass it an array with these two values.
```

scissorsneedfoodtoo:
Unfortunately I'm not able to go through every step at the moment and offer suggestions. All of the information is very good, but could possibly be moved around to other steps to keep each description at about five or six sentences max. Though that's not always possible, we've been trying to keep each test description about that length so there's a similar pace while working through each project.

Looking forward to your next draft!

HKuz:
@scissorsneedfoodtoo thanks for the feedback! Given the number of files, it's understandable you can't go through every one of them up front, but the comments above all make sense. It's been a hectic month, I'll try to get these turned around soon.

scissorsneedfoodtoo:
@HKuz, thank you again for all of your hard work on this project! I had a lot of fun going through it, even if I was only able to get through a handful of lessons.

The end result it amazing and I'm sure learners will have a ton of fun building this project. Please let me know if there's anything I can do to help.

Issue #119, opened 2019-03-08
Learn Basic Algorithms by Building an Algorithmic News Feed


QuincyLarson:

lavanganji:
Hi @QuincyLarson  how can i help in this issue.

picolloo:
I would like to help with this too

QuincyLarson:
@lavanganji @picolloo Both of you have claimed other projects, and I recommend building those first. This one is going to be a real challenge and one or both of you can tackle it a bit later 😄

picolloo:
So I'll choose another one to start

lavanganji:
I will start on this one . @picolloo  you can start on the Shopping Cart

scissorsneedfoodtoo:
Hi @lavanganji, just wanted to check in and see how everything's going with this project. Please post a link to CodePen here when you're finished, and feel free to reach out if you have any questions.

lavanganji:
@scissorsneedfoodtoo  trying to build the pen but not sure where to begin with, any starting suggestion would be very helpful and great.

scissorsneedfoodtoo:
@lavanganji, I imagine this project will go over concepts that are covered in [Basic Data Structures](https://learn.freecodecamp.org/javascript-algorithms-and-data-structures/basic-data-structures) (using JS objects to store key-value pairs, checking properties of an object, and working with arrays), and fall somewhere between the challenges in the [Basic](https://learn.freecodecamp.org/javascript-algorithms-and-data-structures/basic-algorithm-scripting/) and [Intermediate](https://learn.freecodecamp.org/javascript-algorithms-and-data-structures/intermediate-algorithm-scripting) Algorithm Scripting sections in terms of difficulty.

Building a full search engine is probably too difficult for students at this point of the curriculum, so a more accurate title for this project might be "Learn Algorithms by Building a Search Filter". [Here's](https://www.youtube.com/watch?v=3NG8zy0ywIk) a video I found of a book search filter project. You should think of some other data set to use, though, like usernames, email addresses, and phone numbers that can all be searched/filtered.

What do you think, @QuincyLarson? Is that what you had in mind for this project?

lavanganji:
@scissorsneedfoodtoo  out of curiosity i am asking this, is okay i use JAVA as language in JS , this is just a personal preference.

scissorsneedfoodtoo:
[...]

PR #282, opened 2020-01-02
feat: Basic Algorithms

01Clarian:
#[119](https://github.com/freeCodeCamp/CurriculumExpansion/issues/119)  

65 Index pages added for new feature Learn Basic Algorithms.

### DIFF
--- a/basic-algorithms-&-datastructures/index000.js
+++ b/basic-algorithms-&-datastructures/index000.js
@@ -1,0 +1,16 @@
+const newsFeed = {
+}
+
+/*
+Let's learn Basic Algorithms by Building an Algorithmic Newsfeed!
+
+Line one is the empty object newsFeed. Objects have properties which can be set to values. 
+Properties are made up of key : value pairs.
+
+Observe this example of an object: var objectExample = {cats: 4}. 
+
+The object ObjectExample has the key of cats which is set to the the numerical value of 4.
+
+Create a property for Newsfeed with the key of 'page' and set it to the value of 'front_page'.
+*/
+
--- a/basic-algorithms-&-datastructures/index010.js
+++ b/basic-algorithms-&-datastructures/index010.js
@@ -1,0 +1,13 @@
+const newsFeed = {
+    page: "front_page"
+}
+
+/*
+We now have our 'front_page' data property for our newsfeed.
+
+An Object may also contain an array or another object.  For Example: var objectExample = {cats:[]}. *The property cats of objectExample is set
+to an empty array. 
+
+After the property page, create a new articles property and set it to an empty array. 
+*/
+
--- a/basic-algorithms-&-datastructures/index020.js
+++ b/basic-algorithms-&-datastructures/index020.js
@@ -1,0 +1,11 @@
+const newsFeed = {
+    page: "front_page",
+    articles: []
+}
+
+/*
+If we now console newsFeed our newsfeed object reflects its data accordingly: newsFeed page: "front_page"
+
+Now let's add six new empty objects within the 'articles' array.
+*/
+
--- a/basic-algorithms-&-datastructures/index030.js
+++ b/basic-algorithms-&-datastructures/index030.js
@@ -1,0 +1,15 @@
+const newsFeed = {
+    page: "front_page",
+    articles: [{},{},{},{},{},{}]}
+
+/*
+Nested objects are objects within other objects. In the first nested empty object within the articles array, add the following new properties: 
+headline, image, date_published, date_modified, author, likes, dislikes, shares, comments, sections, and url.
+
+Set the property image to an empty array while author should be set to a nested object with the properties username and name. As well, set sections to an object but with the properties name and subscribers. 
+
+*Bonus, initialize all the values of properties that are unnassigned to null.
+
+*/
+
+
--- a/basic-algorithms-&-datastructures/index040.js
+++ b/basic-algorithms-&-datastructures/index040.js
@@ -1,0 +1,33 @@
+const newsFeed = {
+    page: "front_page",
+    articles: [{
+        articles:  [
+            {
+              headline: null,
+              image: [],
+              date_published: null,
+              date_modified: null,
+              author: {
+                username: null,
+                name: null
+              },
+              likes: null,
+              dislikes: null,
+              shares: null,
+              comments: null,
+              section: {
+                name: null,
+                subscribers: null
+              },
+              url: null
+        },{},{},{},{},{}
+    ]}
+]}
+
+/*
+If we refresh and check newsFeed in the console, we can see the object page and the articles array with the nested data along with 5 more empty objects.
+
+Set the username value in the nested object of author to the string "PeoplePerson" and the following name key to the value of the string of "Pat Comb". 
+*/
+
+
--- a/basic-algorithms-&-datastructures/index050.js
+++ b/basic-algorithms-&-datastructures/index050.js
@@ -1,0 +1,31 @@
+const newsFeed = {
+    page: "front_page",
+    articles: [{
+        articles:  [
+            {
+              headline: null,
+              image: [],
+              date_published: null,
+              date_modified: null,
+              author: {
+                username: "PeoplePerson",
+                name: "Pat Comb"
+              },
+              likes: null,
+              dislikes: null,
+              shares: null,
+              comments: null,
+              section: {
+                name: null,
+                subscribers: null
+              },
+              url: null
+        },{},{},{},{},{}
+    ]}
+]}
+
+/*
+Now set the properties likes to the numerical value of 10571 and dislikes to the value of 14.
+*/
+
+
--- a/basic-algorithms-&-datastructures/index060.js
+++ b/basic-algorithms-&-datastructures/index060.js
@@ -1,0 +1,37 @@
+const newsFeed = {
+    page: "front_page",
+      articles: [{
+        headline: "Programming: Expectations vs. Reality",
+        image: [],
+        date_published: "2019-01-05T08:00:00+08:00",
+        date_modified: "2019-01-05T09:20:00+08:00",
+        author: {
+          username: "PeoplePerson",
+          name: "Pat Comb"
+        },
+        likes: 10571,
+        dislikes: 14,
+        shares: 54,
+        comments: 461,
+        section: {
+          name: "Programming",
+          subscribers: 18101
+        },
+        url: "http://www.binary-bugle.com/programming-expectations-vs-reality"
+      },{},{},{},{},{}
+    ]}
+
+/*
+
+Now that we are getting the hang of the structure code of our database, we have autocompleted the rest of database.
+
+To select a specific property in an Object, we can use dot notation. Observe the following object: var learnJavascript = {Today: 'Build an algorithmic newsfeed.', Tomorrow:'Launch new company.'}
+
+We can access the property value of Today in the object learnJavascript with dot notation as follows:
+learnJavascript.Today // "Build an algorithmic newsfeed." 
+
+Access the newsFeed object's property articles in the same manner. 
+
+*/
+
+
--- a/basic-algorithms-&-datastructures/index070.js
+++ b/basic-algorithms-&-datastructures/index070.js
@@ -1,0 +1,33 @@
+const newsFeed = {
+  page: "front_page",
+    articles: [{
+      headline: "Programming: Expectations vs. Reality",
+      image: [],
+      date_published: "2019-01-05T08:00:00+08:00",
+      date_modified: "2019-01-05T09:20:00+08:00",
+      author: {
+        username: "PeoplePerson",
+        name: "Pat Comb"
+      },
+      likes: 10571,
+      dislikes: 14,
+      shares: 54,
+      comments: 461,
+      section: {
+        name: "Programming",
+        subscribers: 18101
+      },
+      url: "http://www.binary-bugle.com/programming-expectations-vs-reality"
+    },{},{},{},{},{}
+  ]}
+
+/*
+To access nested objects we can also use bracket notation and a numerical value to select with object as index. *Note computers count values starting at zero.
+
+For example, in the following object countObject, we can select and change the count value property of 1 to 0 in our first index. 
+var countObject = {nestedCounts: [{count: 1}, {count: 2}]} 
+countObject.nestedCounts[0].count = 0 // returns countObject = {nestedCounts: [{count: 0}, {count: 2}]}
+
+Back in our newsFeed, the author with the username 'PeoplePerson' has now officially changed his name to 'Mr. Coffee'.
+Change the string of 'PeoplePerson' to the string of 'Mr Coffee' accordingly by selecting the username value in the nested author object. 
+*/
--- a/basic-algorithms-&-datastructures/index080.js
+++ b/basic-algorithms-&-datastructures/index080.js
@@ -1,0 +1,29 @@
+const newsFeed = {
+  page: "front_page",
+    articles: [{
+      headline: "Programming: Expectations vs. Reality",
+      image: [],
+      date_published: "2019-01-05T08:00:00+08:00",
+      date_modified: "2019-01-05T09:20:00+08:00",
+      author: {
+        username: "Mr. Coffee",
+        name: "Pat Comb"
+      },
+      likes: 10571,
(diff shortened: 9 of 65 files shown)


vkWeb:
@scissorsneedfoodtoo Can I take a look into this and give some suggestions? 
Actually, I deeply care about the algorithms section because that's what we all do after all.

vkWeb:
@01Clarian @scissorsneedfoodtoo By looking at the issue it looks like you both had a meeting. So it would be great if anyone of you can tell me how you decided to approach this section? 

Here's how I envisioned at approaching this section. I'm writing some points:

- I thought of approaching this in a functional way. There will be some helper functions for sorting, searching, interacting with object values / properties.

- For e.g. we can use selection or any other simple sort in this section and then I can refactor the intermediate section to use quick or merge. We will just tell campers how those sorts work and it will be their job to implement it. 

- And then a lot of small functions that do different tasks like a function to reverse a string, converting to title case and like.

- Introducing those functions in increasing order of difficulty.

- We should not tell campers to do this or that way. We should just tell them what we expect this function to do or return, much like the challenge-based approach but in the end, they will realize how algorithms apply to a real app via this project-oriented approach.

- We should not use `map, filter, reduce, sort` or any shortcut in our solutions. We should also restrict campers from using those. Like if they are writing a sorting algorithm then we should not allow `sort` function in their solution.

- Writing our own versions of those shortcuts will give campers an insight on approaching problems when they don't know an inbuilt function.

01Clarian:
@vkWeb 

Thanks for your thoughts and feedback on this. 

From the original assignment thread and after multiple discussions, my understanding of the idea for this specific assignment is to create a project that introduces practical methods and approaches for organizing, accessing, and modifying a simple newsfeed database while introducing and reviewing basic algorithms and updated methods including ES6. 

In this sense, building up and introducing the data and then bringing in for loops and comparing fundamental structures,  integrating sort, map, filter, and reduce enhances the overall learning experience, not hinders. And to completely shy away from such important tools, although remains a classical educational approach, is not necessarily the most effective way of learning while bridging the gaps. 

When it comes to pure exercises, I wholeheartedly agree with you, however with project based exercises, from my own personal experience, or going through the other projects in the curriculum, having a directional approach (as long as it's written in such a format to provide insight and help maximize muscle practice and getting used to concepts with a level of ease) is still very valuable and effective.

This works from both a point of solidifying your knowledge from the ground up, reviewing and demonstrating not just why they work, but 'how', while providing practical, contextual insight.

I would also suggest that the functions in this index map are introduced from simplicity to complexity. While I agree that a general and open approach to learning is important, keeping this particular project (from my understanding, there already are a few other projects which dive into solely pure basics) which sets a goal towards modifying data completely open ended without some form of a directional learning guide may also hinder the process of learning.  

However, this project does not neglect the basics by any means. That is why it's so intentionally broken down and does not throw map() out as a quick solution at all, but as part of a stepping stone, a piece of the puzzle to the whole and requests careful consideration as well as provides additional exercises and contexts.

Again, contextually speaking, the project's aim is to demonstrate various techniques in a ground up building format which, in my opinion, is not at all affirming one way to do things, but providing insight into the tools and updated approaches that are available. 

On the otherhand, if this project is not going to be based on an objective ie. working with algorithms on a newsfeed, I would tend to agree more with you to take on a more extreme general and purist approach. However, with the objective in mind, I think it's useful and helpful to include those methods and const and let as long as we're explaining things and building as we go. In that sense, it's more of an experiential design as opposed to a classical one.

scissorsneedfoodtoo:
Thank you for all of your hard work here @01Clarian, and great discussion between you and @vkWeb. I haven't had a chance to really dig into the steps yet, but here are some general thoughts. 

Like @vkWeb mentioned, the idea is to have a handful of functions by the end of the project that do things like sort through the newsfeed by author, by date, likes, filter by which sections a user is subscribed to, and so on. Using `map`, `filter`, and `sort` are fine depending on how they're used and introduced. As a whole these new projects will have a lot more guidance than the current challenges, and testing for these small steps often requires us to limit the possible inputs.

I'll take a look now and leave a more detailed review as comments.

basic-algorithms-&-datastructures/index000.js
```diff
@@ -0,0 +1,16 @@
+const newsFeed = {
+}
+
+/*
```
scissorsneedfoodtoo: These projects should have an introduction page, so we can shorten this a bit:

/*
Line one is the empty `newsFeed` object. Objects have properties which can be set to values. Properties are made up of key : value pairs: 

```js
const  objectExample = { cats: 4 } 
```

Create a property for `newsfeed` with the key of `page` and set it to the value of `'front_page'`.
*/

basic-algorithms-&-datastructures/index010.js
```diff
@@ -0,0 +1,13 @@
+const newsFeed = {
+    page: "front_page"
+}
+
+/*
```
scissorsneedfoodtoo: Working with objects will be covered in depth in projects that fall before this one. Review is always good though, and I just tried to shorten the next challenges where learners build the object:

/*
An object may also contain an array or another object. For example: 

```js
const objectExample = { cats:[] }
```

After the property `page`, create an `articles` property and set it to an empty array. 
*/

basic-algorithms-&-datastructures/index020.js
```diff
@@ -0,0 +1,11 @@
+const newsFeed = {
+    page: "front_page",
+    articles: []
+}
+
+/*
```
scissorsneedfoodtoo: /*
Now let's add six new empty objects within the `articles` array.
*/

basic-algorithms-&-datastructures/index030.js
```diff
@@ -0,0 +1,15 @@
+const newsFeed = {
+    page: "front_page",
+    articles: [{},{},{},{},{},{}]}
+
+/*
```
scissorsneedfoodtoo: This step could probably be broken up into two smaller steps. The first could read:

/*
Nested objects are objects within other objects. In the first nested empty object within the articles array, add the following properties: `headline`, `image`, `date_published`, `date_modified`, `author`, `likes`, `dislikes`, `shares`, `comments`, `sections`, and `url`.

Set `image` to an empty array and both `author` and `section` to empty objects. Assign all other properties the value `null`.
*/

And the next step could be:

/*
In the `author` object, add the properties `username` and `name` and set both to the value `null`.

Also in `section`, add the properties `name` and `subscribers` and set them both to `null`.
*/

basic-algorithms-&-datastructures/index040.js
```diff
@@ -0,0 +1,33 @@
...
+              section: {
+                name: null,
+                subscribers: null
+              },
+              url: null
+        },{},{},{},{},{}
+    ]}
+]}
+
+/*
```
scissorsneedfoodtoo: /*
Set the value of `username` in the nested `author` object to the string "PeoplePerson" and the `name` key to the string "Pat Comb".
*/

basic-algorithms-&-datastructures/index050.js
```diff
@@ -0,0 +1,31 @@
...
+              section: {
+                name: null,
+                subscribers: null
+              },
+              url: null
+        },{},{},{},{},{}
+    ]}
+]}
+
+/*
```
scissorsneedfoodtoo: /*
Now set the property `likes` to the number 10571 and dislikes to 14.
*/

basic-algorithms-&-datastructures/index060.js
```diff
@@ -0,0 +1,37 @@
...
+          name: "Programming",
+          subscribers: 18101
+        },
+        url: "http://www.binary-bugle.com/programming-expectations-vs-reality"
+      },{},{},{},{},{}
+    ]}
+
+/*
+
+Now that we are getting the hang of the structure code of our database, we have autocompleted the rest of database.
```
scissorsneedfoodtoo: I'm not sure if `index060.js` and `index070.js` are the best places to review dot and bracket notation. It might be better to include a quick review in a later step when they're actually used to access or set a value, but that's just my quick take.

basic-algorithms-&-datastructures/index070.js
```diff
@@ -0,0 +1,33 @@
...
+      section: {
+        name: "Programming",
+        subscribers: 18101
+      },
+      url: "http://www.binary-bugle.com/programming-expectations-vs-reality"
+    },{},{},{},{},{}
+  ]}
+
+/*
+To access nested objects we can also use bracket notation and a numerical value to select with object as index. *Note computers count values starting at zero.
```
scissorsneedfoodtoo: Building off of my last comment, this could possibly be reworked into a "new username" function later.

basic-algorithms-&-datastructures/index080.js
```diff
@@ -0,0 +1,29 @@
...
+        subscribers: 18101
+      },
+      url: "http://www.binary-bugle.com/programming-expectations-vs-reality"
+    },{},{},{},{},{}
+  ]}
+
+newsFeed.articles[0].author.username = 'Mr Coffee'
+/*
+
+Continue practice using bracket and dot notation with objects by changing the number of subscribers to 18103 and then back to its original value of 18101.
```
scissorsneedfoodtoo: We could remove this lesson. It would be difficult to write a unit test to check for this.

basic-algorithms-&-datastructures/index100.js
```diff
@@ -0,0 +1,142 @@
...
+      },
+      url:
+        "http://www.binary-bugle.com/everything-youve-ever-wanted-to-know-about-ai"
+    }
+  ]
+};
+
+let newsFeedArray = []
+
+/*
```
scissorsneedfoodtoo: This is a very good explanation, but I believe the example code is missing the last curly brace. Here's a quick edit with an updated variable name:

/*
The `for...in` statement iterates over the properties of an object keyed by strings. For example: 

```js
const objExample = {animals: [{cats: 3}, {dogs: 4}, {birds: 7}]}

let animalsArray = [];

for (props in objExample) {
  animalsArray = objExample[props];
}
```

Convert our object newsFeed into an array by iterating over it with the `for...in` statement
and returning the result of each iteration into our newsFeedArray array. 
*/

Also, sorry, I believe I missed this in one of your earlier prototypes, but would it be easier to do the following?: `const newsFeedArray = newsFeed.articles;`

basic-algorithms-&-datastructures/index630.js
```diff
@@ -0,0 +1,500 @@
...
+// Starred Artciles Feature
+for(char of newsFeedArray) {
+ if(char.likes > 900) {
+  char.starred = true
+ } else {
+  char.starred = false
+ }
+}
+
+/*
```
scissorsneedfoodtoo: I wasn't able to go over every step in detail, but I think this is off to a great start. I like the way you structured the lessons here, and think that each one builds off the last quite nicely. 

One thing we can do in the next draft is to remove the console.log statements. Prompting learners to write one and check the output is great, but the console can get cluttered really quickly. Also I see what you're doing with things like the hard coding of `filter` to return just the articles by 'Julie Holmes', then moving on to a more flexible solution. That's great, and when that happens don't be afraid to prompt learners to delete what they already wrote before refactoring.

When testing the final code I noticed that the `newsFeed` array seems like it has some repeats, like the article "Will Web Development Ever Rule the World?". Some articles have the `id` property while others don't, and there seems to be fewer articles than in the CodePen link. In the final version of the project we can use separate files for the `newsFeed` object, and use `export` and `import` to make it accessible from the main file.

Another thing I noticed is the `prompt` method. For this project I don't think we need a UI, and I picture it using just the code editor and console. We could probably remove the HTML page entirely and rework that code into a function that takes a string as a parameter and returns that author's articles.

vkWeb:
@scissorsneedfoodtoo @01Clarian You have put some good work here but IMO this will not give campers a chance to put their own mind into the code - which should be our aim here. 

This is more like "you need to do this then that...". It should be "here's what we expect, you need to figure out the 'how'...". 

Anyways, thanks for contributing @01Clarian :) Let's see how things turn out.

vkWeb:
@scissorsneedfoodtoo I'll push the intermediate section soon.

basic-algorithms-&-datastructures/index630.js
01Clarian: @scissorsneedfoodtoo Thanks for your consideration and these notes! Very cool to continue building on this and looking forward to further shaping it. 

Just reviewing through your notes and yes it all makes sense and these are all improvement I can take a look at. Just from quickly going over things this morning -  Noted regarding the console.logs and the removing code prompts while refactoring to keep things more clean overall. 

I remember you mentioning the possibility of importing and exporting the newsfeed data earlier on our meeting as well and I think that would also create more organization and clarity for  project interaction as well as perhaps provide a more 'realistic' experience.

Can also make that change to the final prompt and reshape the code at the end bearing in mind a more absent approach to the UI including removing the HTML page. Makes sense.

Will go through all the other specific lesson notes in detail on the second draft as well as some proof reading and looking forward to getting something improved over to you soon.

Thanks!

01Clarian:
@vkWeb Thanks again for your thoughts and initial feedback with this. 

I also do  appreciate the delicate nature of the learning process and how we develop these skills and respect the passion and vigor that you have expressed.  

As I mentioned above, in my own personal experience, having advanced my own problem solving as well as technical skills over the years with software and hardware engineering (mostly with self learning), some of the most important 'boosters' or moments where I felt like I advanced to the next level were from working from start to finish on a project with a leader who would share and direct our patterns and approaches. Now, that being said, everyone's personalities and methods of expressing themselves are always different, right? 

But my point is not about how well did they stop and mentor along the way, as much as what stuck with me in the very context of building something and having a map of knowledge to follow is a very valuable learning tool in itself. 

Often times, in the project, following along the workflow of someone from start till completion provided a deeper context into the mechanism of our workflow and structure as a whole. There is deep quality and value towards working on projects and following and discovering patterns along the way can lead to many aha moments and create an invaluable combination of memory mapping as well as motivation. That's why in part, I feel this learning approach has recently taken such popularity with online courses in general. However, that being said, that alone, is not enough and there always caveats and traps to be weary of as with all things. And as long as we're mindful in all of our approaches, including classical open ended exercises, we can create a valuable learning platform.

But I think that having that system, a project system as a core structure, with practical and fresh approaches to learning which do not neglect the basics, and challenge the learner to question and review the material on their own is awesome. It emulates that special learning experience of working on real life projects and those lessons can definitely open your mind while providing that backbone to teach and solidify core concepts.

So all of that being said, appreciate the dialog here and your input. Looking forward to continue developing this checking out the intermediate one.

scissorsneedfoodtoo:
@01Clarian, thank you again for all of your work on this. Lots of great discussion between you and @vkWeb here, and a lot for us to think about going forward.

Before we continue to break this down into steps, would you be able to work on a final prototype in  CodePen or CodeSandbox? CodeSandbox should allow us to keep the article and user objects in separate files and import them into other parts of the project. Then we'll be able to really flesh out all the functionality before really breaking things down into steps.

A lot of your descriptions look good, so I'm sure we'll be able to use them in the current steps or move them to whatever new steps come up in the future.

01Clarian:
@scissorsneedfoodtoo, Sounds good. I'll work on the final prototype on CodeSandbox and separate the files. I'm leaving to India for a week, but will find some time coming up soon to get this done. 

Thanks

vkWeb:
@01Clarian Hey! I live in India. Which city are you planning to visit?

01Clarian:
@vkWeb Working in Delhi, Bombay, and Bangalore. 

Actually, looking at organizing with a friend Software Developer in Mumbai a meetup for self learners and web developers over there. If you are located there, would be great. It's the following Saturday.

In any case, let me know your e-mail or best way to contact and I can send over more information.

scissorsneedfoodtoo:
@01Clarian, awesome, thank you for the updates. CodeSandbox sounds great, I'm a fan of the platform. 

Hope you're having a great trip to India! Would be cool if you and @vkWeb were able to meet up there :)

vkWeb:
@01Clarian Wow, great. That sounds cool. 

I live in a small city called Raipur. It's too far away from those cities. But in the future, I'll definitely plan to visit & work with devs in those cities. 

Here's my email: `vivekmittalagrawal@gmail.com` in case you need it for your future visits :) 

Btw, how was your experience? Did you face any difficulties?

Issue #120, opened 2019-03-08
Learn Data Structures by Building a Shopping Cart


QuincyLarson:

picolloo:
I'd like to help with this project

lavanganji:
@QuincyLarson  could you guide us to take part in this project

QuincyLarson:
@picolloo @lavanganji Which of you is interested in tackling this one? This should be a fun one!

picolloo:
I guess we can cooperate, I don't mind to split work

picolloo:
I guess we can cooperate, I don't mind to split work

picolloo:
I can leave  this one to other contribuitors and comeback later if it's necessary

ritute:
Hi, I'm interested in building an example project for this!

QuincyLarson:
@ritute OK - great! See if you can get an example project together that we can then discuss before you start creating tests for it.

scissorsneedfoodtoo:
Hi @ritute, just wanted to check in and see how everything's going with this project. Please post a link to CodePen when you're finished, and feel free to reach out if you have any questions.

scissorsneedfoodtoo:
@ritute, any progress on your prototype?

shimphillip:
@scissorsneedfoodtoo  Would love to take this!

I noticed you are in Korea, are you working as a dev? I am from Korea :)

scissorsneedfoodtoo:
@shimphillip, 안녕하세요? We would love your help on this project! You can read @QuincyLarson's [announcement](https://www.freecodecamp.org/forum/t/help-us-build-version-7-0-of-the-freecodecamp-curriculum/263546) about the new curriculum, along with some [FAQs](https://www.freecodecamp.org/forum/t/freecodecamp-curriculum-7-0-faqs/269346) I wrote. Hopefully those will help you get started, but feel free to reach out if you have any questions. The first step is to build a prototype on a platform like CodePen, CodeSandbox, or Glitch, then post it here for review. Then you can start breaking the project down into steps like the ones submitted as Pull Requests.

Yes, I'm still in Korea. I came here as an English teacher, and now I'm working remotely from Daegu. Where in Korea are you from? I see you're in Austin now. I've always wanted to visit there!

shimphillip:
반가워요! Yes, that should be enough to get me started.

Great! I figured you went to Korea to teach English initially. Austin is great! Supposedly one of the best places to live. Definitely a place to checkout.

[...]

PR #292, opened 2020-03-26
initial draft

hassaanp:
This is an initial draft to demonstrate how we can introduce OOP organically in the shopping cart project.

As per our call @scissorsneedfoodtoo, kindly review and comment if you think I should proceed further in this manner or if the approach needs adjustment.

Thanks

Links #120

### DIFF
--- a/learn-data-structures-by-building-a-shopping-cart/index07.js
+++ b/learn-data-structures-by-building-a-shopping-cart/index07.js
@@ -22,7 +22,7 @@ const products = [
 ];
 
 /*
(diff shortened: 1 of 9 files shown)


learn-data-structures-by-building-a-shopping-cart/index07.js
```diff
@@ -21,24 +21,21 @@ const products = [
...
-  sayName (name, list) {
-
-  }
-  ```
-  */
-
-  addItem() {
-
+/*
+You can also add methods to an object. If a function referenced by a property of the object, we call it a method of the object.
```
hassaanp: typo: if a function **is** referenced

PR #295, opened 2020-04-07
introducing concepts of object generators + working to Object.create

hassaanp:
This PR builds on the previous modifications.
I have touched on how an existing object can be reused and highlighted the trivial copy-paste method. This organically moves on towards introducing the idea of using functions to build objects to make code reusable and maintainable.
In the end, I have highlighted the problem of using the object generator function which forced me to give a rough idea of what memory is.

I have tried to keep it as digestible as I could but I understand that this may require several iterations.

@scissorsneedfoodtoo would love to hear your thoughts.

### DIFF
--- a/learn-data-structures-by-building-a-shopping-cart/index11b.js
+++ b/learn-data-structures-by-building-a-shopping-cart/index11b.js
@@ -42,6 +42,15 @@ const primary = shoppingCartGenerator();
 const wishlist = shoppingCartGenerator();
 
 /*
-    There is a way of reusing pieces of code in memory to mitigate the problem we just saw.
-    Inst
+  A smart approach would be to move the functions `getItems` and `addItem` to an object that can be used
+  to store all the methods we wish our shoppingCart objects to have access to.
+
+  ```
+  const functionStore = {
+    method1 = ()=>{
+      console.log('hey, I am a function')
+    }
+  }
+  ```
+  Create an empty object called `functionStore`.
 */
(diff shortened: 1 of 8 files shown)


learn-data-structures-by-building-a-shopping-cart/index11b.js
```diff
@@ -0,0 +1,56 @@
...
+    return shoppingCart.items;
+  };
+  return newShoppingCart;
+}
+
+const primary = shoppingCartGenerator();
+const wishlist = shoppingCartGenerator();
+
+/*
+  A smart approach would be to move the functions `getItems` and `addItem` to an object that can be used
```
scissorsneedfoodtoo, line 45: Your explanations in the last two steps were very good. Unfortunately they're a bit too long, and there are no code changes for us to test. If possible, we should shorten the explanation to something like the following:

Unfortunately we have a problem -- every time an object is created using `shoppingCartGenerator`, we are creating copies of the object's code in memory.
Inside each shopping cart, we have separate copies of the `items` property, `addItem` method and `getItems` method.
If we add more methods and create more shopping carts, memory will be used up more quickly and limit the performance of our computer/browser.
A smart approach would be to move the functions `getItems` and `addItem` to an object that can be used to store all the methods we wish our `shoppingCart` objects to have access to:
```
const functionStore = {
  method1: () => {
    console.log('Hey, I am a function')
  }
}
```
Create an empty object called `functionStore`.

scissorsneedfoodtoo:
@hassaanp, thank you again for all of your hard work on this. Your new steps are coming along great and I'm looking forward to your next PR :+1: :+1:

PR #312, opened 2021-10-21
feat: penguin draft

nhcarrigan:
Checklist:

- [X] My pull request has a descriptive title (not a vague title like `Update index.md`)

Closes #XXXXX

For our second-to-last project!

### DIFF
--- a/css-penguin/index.html
+++ b/css-penguin/index.html
@@ -10,9 +10,8 @@
     <div class="sun"></div>
     <div class="penguin">
       <div class="shirt">
-        <div>I</div>
-        <div class="heart">💜</div>
-        <div>CSS</div>
+        <div>💜</div>
+        <p>I CSS</p>
       </div>
       <div class="penguin-bottom">
         <div class="right-hand"></div>
--- a/css-penguin/index.html
+++ b/css-penguin/index.html
@@ -1,0 +1,59 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <link rel="stylesheet" href="./styles.css" />
+    <title>CSS Penguin</title>
+  </head>
+  <body>
+    <div class="left-mountain"></div>
+    <div class="background-mountain"></div>
+    <div class="sun"></div>
+    <div class="penguin">
+      <div class="penguin-bottom">
+        <div class="right-hand"></div>
+        <div class="left-hand"></div>
+        <div class="right-feet"></div>
+        <div class="left-feet"></div>
+      </div>
+      <div class="penguin-top">
+        <div class="right-cheek"></div>
+        <div class="left-cheek"></div>
+        <div class="belly"></div>
+        <div class="right-eye">
+          <div class="sparkle"></div>
+        </div>
+        <div class="left-eye">
+          <div class="sparkle"></div>
+        </div>
+        <div class="blush-right"></div>
+        <div class="blush-left"></div>
+        <div class="beak-top"></div>
+        <div class="beak-bottom"></div>
+      </div>
+    </div>
+    <div class="penguin shift-right">
+        <div class="penguin-bottom">
+          <div class="right-hand"></div>
+          <div class="left-hand"></div>
+          <div class="right-feet"></div>
+          <div class="left-feet"></div>
+        </div>
+        <div class="penguin-top">
+          <div class="right-cheek"></div>
+          <div class="left-cheek"></div>
+          <div class="belly"></div>
+          <div class="right-eye">
+            <div class="sparkle"></div>
+          </div>
+          <div class="left-eye">
+            <div class="sparkle"></div>
+          </div>
+          <div class="blush-right"></div>
+          <div class="blush-left"></div>
+          <div class="beak-top"></div>
+          <div class="beak-bottom"></div>
+        </div>
+      </div>
+    <div class="ground"></div>
+  </body>
+</html>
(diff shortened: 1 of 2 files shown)


nhcarrigan:
@ahmadabdolsaheb @scissorsneedfoodtoo Okay I've given this another look.

I've removed the second penguin, as I no longer need it (it was there to teach `scale`).

I've given the first penguin a "shirt", which allows reinforcement of the `::before` pseudo selector.

I've used `float` to give the shirt the I heart css text, with some additional properties to position things a bit better

I've used `:hover` to give the penguin a transition where it expands on hover, which is where `scale` is now covered.

scissorsneedfoodtoo:
Hey @nhcarrigan, thanks for your patience.

I took a look at the updates and everything looks great. It seems like all the necessary properties are covered with your latest changes.

I just noticed a couple of minor things I'll leave as review comments. They may be due to my screen and resolution scaling, though, so feel free to ignore them if they look good on your end.

css-penguin/index.html
```diff
@@ -0,0 +1,41 @@
+<!DOCTYPE html>
+<html>
+  <head>
+    <link rel="stylesheet" href="./styles.css" />
+    <title>CSS Penguin</title>
```
scissorsneedfoodtoo, line 5: There seems to be some margin collapsing on smaller screens:

image

But setting the viewport meta tag seems to fix this:

```suggestion
    <title>CSS Penguin</title>
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
```

ahmadabdolsaheb:
Screen Shot 2021-11-08 at 2 39 17 PM


I played around it a little bit. Here are a couple of thoughts:
1) It might be simpler to have a div and a p inside the shirt element: https://codepen.io/abdolsa/pen/mdMLWNj
2) It seems like we should not be introducing the animation properties in this challenge based on the lesson plan sheets.
3) The background color template could be improved, I changed a couple of things and added them to this pin: https://codepen.io/abdolsa/pen/wvqjWPd?editors=1100

Notes regarding the last shared pin:
1)  There are color vars in some places, hex, rgb, and css color names at others and we could clean up and  introduce some consistency.
2) Since the t-shirt cover a large part of the torso, I just replaced them with the shirt.
3) If the t-shirt does not look good enough, we could make a smaller a little bit.

scissorsneedfoodtoo:
@ahmadabdolsaheb, wow, that pen looks great. I really like the new background colors -- it feels like a more natural environment for a penguin.

Also, that's a good point about not introducing animation properties in this project based on the sheet. The Ferris wheel project is much simpler overall, and focuses only on animation.

Maybe the easiest thing to do would be to switch the order of the projects -- introduce basic animations with the Ferris wheel first, then review animations with this penguin project and cover more advanced animation topics like `transition` to control animation speed and timing.

ahmadabdolsaheb:
Thanks Kris. Also the hover state is difficult to see if a user is developing using their mobile phone.

nhcarrigan:
Alright I've pulled over the latest changes - I did have to position the text a bit more explicitly, but it's looking better now!

scissorsneedfoodtoo:
This is looking good!

About the `:hover` pseudo element being difficult for learners on mobile devices, I think you're right. It would probably be better to use `:active` instead.

With a mouse, `:active` would cause the penguin to scale up while the mouse button is held down, and scale down when the mouse button is released. It should work similarly on touch screens, where the penguin scales up while pressed, and scales down when the touch event ends.

Other than that, I believe this should be ready to break up into steps.

nhcarrigan:
Alright, I'll switch that selector to active and start breaking into steps in the morning, unless @ahmadabdolsaheb has additional feedback.

ahmadabdolsaheb:
Here is another round of cleaning up:https://codepen.io/abdolsa/pen/yLoqQBO
Notes:
1) if we remove the float: right on the heart, not much happens
2) if the elements are ordered from the background to foreground  and written from top to bottom in html, we could lose a lot of the z-index properties.

As Kris mentioned, it might make sense to have this as the last project. Since the the animated Pinguin was the last challenge on the basic CSS, the waving of the Pinguin was purposefully designed as a farewell.   :)

nhcarrigan:
I don't see an issue with switching this and the Ferris wheel so the penguin comes last 🙂

scissorsneedfoodtoo:
Took another look and everything LGTM :+1: :+1: 

Ahmad brings up a good point about `float` and the order of elements. It's true that removing `float` doesn't change much. But it may be one of those things that's more obvious when building the project.

And for `z-index` that could be something to mention while building the project -- that with some refactoring, that it would be possible to remove them. Otherwise I don't see a problem with keeping them in.

ShaunSHamilton (approved):
This LGTM 👍 

I agree with Nich about the `z-index`; this is probably the perfect opportunity to make use of such a property and not worry about how it might affect accessibility. Besides, other projects mention the importance of HTML structure for an accessible of control flow.

nhcarrigan:
Awesome thanks for the second set of eyes @ShaunSHamilton - I think we're good to move forward with this one. Did you still want to tackle breaking it down into steps?

ShaunSHamilton:
> Did you still want to tackle breaking it down into steps?

I think so, yes. I will get on it tomorrow 👍

nhcarrigan:
All yours! 💜

PR #314, opened 2022-01-04
refactor: use tables

nhcarrigan:
Checklist:

Closes #XXXXX

Discussion with @ShaunSHamilton regarding https://github.com/freeCodeCamp/freeCodeCamp/issues/44609 led to a suggestion to remove `tabindex`, and also convert this project to use HTML tables (which we do not teach elsewhere).

### DIFF
--- a/css-balance-sheet/index.html
+++ b/css-balance-sheet/index.html
@@ -5,105 +5,117 @@
     <title>freeCodeCamp Balance Sheet</title>
     <link rel="stylesheet" type="text/css" href="./styles.css" />
   </head>
-  <body>
-    <table id="sheet">
-      <caption>freeCodeCamp's annual budget</caption>
-      <thead>
-        <tr id="header">
-          <th>Balance Sheet</th>
-        </tr>
-        <tr id="subheader">
-          <th>freeCodeCamp</th>
-        </tr>
-      </thead>
-      <tbody>
-        <tr id="years">
-          <td>2019</td>
-          <td>2020</td>
-          <td class="current">2021</td>
-        </tr>
-        <tr>
-          <th>Assets</th>
-        </tr>
-        <tr class="data">
-          <td class="name">Cash</td>
-          <td>$25</td>
-          <td>$30</td>
-          <td class="current">$28</td>
-        </tr>
-        <tr>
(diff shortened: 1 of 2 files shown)


ShaunSHamilton (changes requested):
Thank you, for tackling this, Nich.

I am going to be a pain and suggest some accessibility improvements - otherwise @bbsmooth will _be on our case_ 😉 

---

1) Give the table a `caption` element describing the table
2) Move the `Balance Sheet` and `freeCodeCamp` into their own `thead` element
3) Explicitly define where the `tbody` should be (otherwise HTML tries to guess where to put it)
4) OPTIONAL _because it is relatively advanced_ : Include `headers` and `scope` attributes for the `th` and `td` elements to relate essentially what is a tag/heading and what is a value.

nhcarrigan:
@ShaunSHamilton for some reason I had it in my head that you said you'd look at this. 🤦🏻‍♀️ 

Anyway, I've applied 1 2 and 3. 4 doesn't seem like it would fit well with the table structure, which *might* mean it's not a good table? 🤔 Also, any thoughts on how we want to style the caption?

ShaunSHamilton:
> for some reason I had it in my head that you said you'd look at this. 🤦🏻‍♀️

I think I did say this (on Chat, or something), and I did look into it, but never found a particularly nice way of doing things.

> which might mean it's not a good table?

It is an odd table.

How about, before we waste our time implementing stuff that might not be correct, we get another opinion? I will post this on the forum.

> Also, any thoughts on how we want to style the caption?

I am not sure about the accessibility of such, but I always prefer captions at the bottom:

```html
<caption style="caption-side:bottom">Caption describing table</caption>
```

nhcarrigan:
> How about, before we waste our time implementing stuff that might not be correct, we get another opinion? I will post this on the forum.

This is not a bad idea, tbh.

ShaunSHamilton (approved):
Well... LGTM 👍

nhcarrigan:
Gave this one final look. I think we might be missing a selector or two, but IMHO it's much more beneficial to teach good accessibility practises and HTML tables.

nhcarrigan:
Will break down into steps tomorrow

PR #315, opened 2022-07-15
feat: stat calculator

naomi-lgbt:
Checklist:

Closes #XXXXX

Small project to teach array methods before diving in to the functional programming spreadsheet

### DIFF
--- a/learn-array-methods-by-building-a-statistics-calculator/index.html
+++ b/learn-array-methods-by-building-a-statistics-calculator/index.html
@@ -16,9 +16,21 @@
       <input type="text" name="numbers" id="numbers" />
       <button type="submit">Calculate</button>
     </form>
-    <div class="results"q>
+    <div class="results">
+      <p>
+        The <dfn>mean</dfn> of a list of numbers is the average, calculated by
+        taking the sum of all numbers and dividing that by the count of numbers.
+      </p>
       <p class="bold">Mean: <span id="mean"></span></p>
+      <p>
+        The <dfn>median</dfn> of a list of numbers is the number that appears in
+        the middle of the list, when sorted from least to greatest.
+      </p>
       <p class="bold">Median: <span id="median"></span></p>
+      <p>
+        The <dfn>mode</dfn> of a list of numbers is the number that appears most
+        often in the list.
+      </p>
       <p class="bold">Mode: <span id="mode"></span></p>
     </div>
   </body>
--- a/learn-array-methods-by-building-a-statistics-calculator/index.html
+++ b/learn-array-methods-by-building-a-statistics-calculator/index.html
@@ -1,0 +1,24 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <link rel="stylesheet" href="./styles.css" />
+    <script src="./script.js"></script>
+    <title>Statistics Calculator</title>
+  </head>
+  <body>
+    <h1>Statistics Calculator</h1>
+    <p>Enter a list of comma-separated numbers.</p>
+    <form onsubmit="calculate(); return false;">
+      <label for="numbers">Numbers:</label>
+      <input type="text" name="numbers" id="numbers" />
+      <button type="submit">Calculate</button>
+    </form>
+    <div class="results"q>
+      <p class="bold">Mean: <span id="mean"></span></p>
+      <p class="bold">Median: <span id="median"></span></p>
+      <p class="bold">Mode: <span id="mode"></span></p>
+    </div>
+  </body>
+</html>
--- a/learn-array-methods-by-building-a-statistics-calculator/index.html
+++ b/learn-array-methods-by-building-a-statistics-calculator/index.html
@@ -10,6 +10,7 @@
   <body>
     <h1>Statistics Calculator</h1>
     <p>Enter a list of comma-separated numbers.</p>
+    <!-- Teach here that return false prevents a refresh-->
     <form onsubmit="calculate(); return false;">
       <label for="numbers">Numbers:</label>
       <input type="text" name="numbers" id="numbers" />
--- a/learn-array-methods-by-building-a-statistics-calculator/index.html
+++ b/learn-array-methods-by-building-a-statistics-calculator/index.html
@@ -32,6 +32,23 @@
         often in the list.
       </p>
       <p class="bold">Mode: <span id="mode"></span></p>
+      <p>
+        The <dfn>range</dfn> of a list of numbers is the difference between the
+        largest and smallest numbers in the list.
+      </p>
+      <p class="bold">Range: <span id="range"></span></p>
+      <p>
+        The <dfn>variance</dfn> of a list of numbers measures how far the values
+        are from the mean, on average.
+      </p>
+      <p class="bold">Variance: <span id="variance"></span></p>
+      <p>
+        The <dfn>standard deviation</dfn> of a list of numbers is the square
+        root of the variance.
+      </p>
+      <p class="bold">
+        Standard Deviation: <span id="standardDeviation"></span>
+      </p>
     </div>
   </body>
 </html>
--- a/learn-array-methods-by-building-a-statistics-calculator/script.js
+++ b/learn-array-methods-by-building-a-statistics-calculator/script.js
@@ -1,4 +1,5 @@
 const getMean = (array) => {
+  // teach the reduce method
   const mean = array.reduce((acc, el) => acc + el, 0) / array.length;
   return mean;
 };
@@ -10,6 +11,7 @@ const getMedian = (array) => {
 
 const getMode = (array) => {
   const counts = {};
+  // teach the forEach method??
   array.forEach((el) => {
     counts[el] = (counts[el] || 0) + 1;
   });
@@ -22,7 +24,12 @@ const getMode = (array) => {
 const calculate = () => {
   const value = document.querySelector("#numbers").value;
   const array = value.split(/,\s*/g);
+  // teach here that inputs are always strings
+  // also teaching the map method
   const numbers = array.map((el) => parseInt(el));
+  // explain that values which don't parse to a number are NaN
+  // and NaN can't be compared to anything
+  // also teaching the filter method
   const filtered = numbers.filter((el) => !isNaN(el));
 
   const mean = getMean(filtered);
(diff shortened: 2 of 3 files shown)


learn-array-methods-by-building-a-statistics-calculator/index.html
```diff
@@ -0,0 +1,25 @@
...
+  <body>
+    <h1>Statistics Calculator</h1>
+    <p>Enter a list of comma-separated numbers.</p>
+    <!-- Teach here that return false prevents a refresh-->
+    <form onsubmit="calculate(); return false;">
+      <label for="numbers">Numbers:</label>
+      <input type="text" name="numbers" id="numbers" />
+      <button type="submit">Calculate</button>
+    </form>
+    <div class="results"q>
```
jdwilkin4, line 19: It looks like we have a stray "q" here

learn-array-methods-by-building-a-statistics-calculator/styles.css
```diff
@@ -0,0 +1,18 @@
...
+  margin: 0;
+  background-color: rgb(27, 27, 50);
+  text-align: center;
+  color: #fff;
+}
+
+button {
+  background-color: rgb(59, 59, 79);
+  border: 3px solid white;
+  color: white;
```
jdwilkin4, line 11: I think it would be nice to add the `cursor:pointer;` here

learn-array-methods-by-building-a-statistics-calculator/script.js
```diff
@@ -0,0 +1,42 @@
...
+
+const getMode = (array) => {
+  const counts = {};
+  // teach the forEach method??
+  array.forEach((el) => {
+    counts[el] = (counts[el] || 0) + 1;
+  });
+  const highest = Object.keys(counts).sort(
+    (a, b) => counts[b] - counts[a]
+  )[0][0];
```
jdwilkin4, line 20: I think this is the first time we introduce method chaining.
So we do want to make sure to explain it here. 
Also, this might be the first time we introduce `Object.keys`

learn-array-methods-by-building-a-statistics-calculator/script.js
```diff
@@ -0,0 +1,42 @@
...
+};
+
+const getMedian = (array) => {
+  const median = array[Math.floor((array.length - 1) / 2)];
+  return median;
+};
+
+const getMode = (array) => {
+  const counts = {};
+  // teach the forEach method??
```
jdwilkin4, line 14: I think it is fine to teach it.
The `forEach` method might have come up earlier, but this project is small enough were we can teach it/review it here too.

learn-array-methods-by-building-a-statistics-calculator/index.html
```diff
@@ -0,0 +1,25 @@
...
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <link rel="stylesheet" href="./styles.css" />
+    <script src="./script.js"></script>
+    <title>Statistics Calculator</title>
+  </head>
+  <body>
+    <h1>Statistics Calculator</h1>
+    <p>Enter a list of comma-separated numbers.</p>
```
jdwilkin4, line 12: We might consider adding a short definition for mean, median and mode since some campers might not know or remember what they are.

Maybe we can add another p element or just explain it in the steps.

learn-array-methods-by-building-a-statistics-calculator/index.html
```diff
@@ -0,0 +1,25 @@
...
+    <!-- Teach here that return false prevents a refresh-->
+    <form onsubmit="calculate(); return false;">
+      <label for="numbers">Numbers:</label>
+      <input type="text" name="numbers" id="numbers" />
+      <button type="submit">Calculate</button>
+    </form>
+    <div class="results"q>
+      <p class="bold">Mean: <span id="mean"></span></p>
+      <p class="bold">Median: <span id="median"></span></p>
+      <p class="bold">Mode: <span id="mode"></span></p>
```
jdwilkin4, line 22: It doesn't look like the `bold` class is being used in the CSS. maybe we can get rid of it.

jdwilkin4 (changes requested):
I tested it out locally, and I think it will be a good project to lead into the spreadsheet project. 👍

learn-array-methods-by-building-a-statistics-calculator/index.html
naomi-lgbt: That's the new `q` attribute. You add it to an element when you have no idea what you're doing, like me 🙃

learn-array-methods-by-building-a-statistics-calculator/script.js
naomi-lgbt: I was also considering chaining the `map` and `filter` and explaining it there.

learn-array-methods-by-building-a-statistics-calculator/script.js
naomi-lgbt: We could also chain the `map` and `filter` and explain chaining there, so that we can explain `Object.keys` here.

jdwilkin4:
I updated the spreadsheet to include all of the concepts taught. 👍

scissorsneedfoodtoo:
Thanks for your patience @naomi-lgbt. I took a look and everything looks great. Just noticed a few things, most of which are just things we can discuss more or keep in mind when it's time to break this up into steps.

I'll leave some review comments now.

learn-array-methods-by-building-a-statistics-calculator/styles.css
scissorsneedfoodtoo: I was thinking this too, but when I did some digging, it seems like using `cursor: pointer` is only used for links by default.

It might be better to leave the cursor the way it is, and to use a hover effect over the button since it's styled.

But I have no strong feelings either way. I'm pretty sure that some of the other projects I worked on have a modified cursor for buttons.

I think as long as we're pretty consistent throughout the projects that either way is fine.

learn-array-methods-by-building-a-statistics-calculator/script.js
```diff
@@ -0,0 +1,42 @@
+const getMean = (array) => {
+  // teach the reduce method
+  const mean = array.reduce((acc, el) => acc + el, 0) / array.length;
```
scissorsneedfoodtoo, line 3: I'm sure you're already thinking along these lines, but since this is the first time learners will use the `reduce` method, it might be good to teach the longer form of the method where the accumulator is returned, then to refactor to this more concise version.

learn-array-methods-by-building-a-statistics-calculator/index.html
scissorsneedfoodtoo: Was going to point out the same thing. Could also create a `bold` class and change the `font-weight` for the results.

learn-array-methods-by-building-a-statistics-calculator/index.html
scissorsneedfoodtoo: I agree. Wouldn't hurt to have a short definition as another `p` element, or to use a `span` and a `title` attribute as a tooltip.

learn-array-methods-by-building-a-statistics-calculator/script.js
scissorsneedfoodtoo: Agreed. This would be a great place to review the `forEach` method. Currently it's being introduced in the Music Playlist project.

It would be good to review things like how the `forEach` method is great when you want to call a function for each element in an array, not when you want to return something like with `map` or `reduce`, and that `forEach` returns `undefined`.

learn-array-methods-by-building-a-statistics-calculator/script.js
scissorsneedfoodtoo: > We could also chain the `map` and `filter` and explain chaining there, so that we can explain `Object.keys` here.

That sounds like a great idea.

learn-array-methods-by-building-a-statistics-calculator/styles.css
scissorsneedfoodtoo: I was thinking this too, but when I did some digging, it seems like using `cursor: pointer` is only used for links by default.

It might be better to leave the cursor the way it is, and to use a hover effect over the button since it's styled.

But I have no strong feelings either way. I'm pretty sure that some of the other projects I worked on have a modified cursor for buttons.

I think as long as we're pretty consistent throughout the projects that either way is fine.

naomi-lgbt:
Alright, I've made tweaks based on the above comments, and also added some additional calculations to create opportunities for more advanced array methods (and a nice tie-in to our foundational math course)

learn-array-methods-by-building-a-statistics-calculator/styles.css
bbsmooth: The mouse hover over non-links issues comes up quite a bit in accessibility. By default the browser only uses the pointer for links. I believe this mimics native desktop apps. And then you would use some other styling to indicate the hover state for buttons and such. Some people think this is the way it should always be and others argue that web applications aren't native applications and there can be different rules. 

Personally, I don't consider it a sin to change the mouse to pointer on hover over buttons and such, but I don't usually do that unless it is specifically ask for. So I wouldn't go out of the way to make a point that you need to change the mouse cursor on hover over non-links, because you don't. Leaving the browser defaults is perfectly fine. But if changing the mouse cursor adds something to this then sure, leave it in. I doubt you would ever be docked points in an accessibility audit for it.

jdwilkin4 (approved):
I played around with it locally and I think it looks good 👍

jdwilkin4:
I will also add a checklist of concepts covered for us to keep track of

- [x] arrow functions
- [x] function parameters 
- [ ] reduce method
- [ ] arr.length
- [ ] sort method with numbers
- [ ] Math.floor
- [ ] const and let
- [ ] object literals
- [ ] forEach method
- [ ] OR operator
- [ ] Object.keys
- [ ] accessing elements from an array
- [ ] method chaining
- [ ] parseInt
- [ ] Math.min and Math.max
- [ ] function calls
- [ ] exponentiation operator 
- [ ] Math.sqrt
- [ ] querySelector
- [ ] split method
- [ ] basic regex
- [ ] map method
- [ ] filter method
- [ ] ! operator
- [ ] isNaN
- [ ] innerHTML
- [ ] toFixed

learn-array-methods-by-building-a-statistics-calculator/styles.css
scissorsneedfoodtoo: Awesome, thanks for confirming that @bbsmooth. I wasn't able to find a definitive answer either way, but it's great to know that changing the pointer type for buttons isn't a bad thing accessibility wise.

I believe the other JS projects all use pointers for buttons, so we may stick with that to keep things consistent. If we get enough feedback later saying that we should stick to browser defaults we can do that down the road.

scissorsneedfoodtoo (approved):
Everything LGTM, too. All the new stats look great. Let's get this in.

PR #320, opened 2022-07-29
feat: updating advanced dice game

jdwilkin4:
I found this project in the repo and there are a lot of great concepts taught in this game.
The original contributor did a great job with the prototype.
I made some edits to the CSS and JS. 

Live link: https://3im5gn.csb.app/

## Topics covered

- [x] `querySelectorAll`
- [x] `getElementById`
- [x]  Review of the `Selection Sort` algorithm 
- [x] arrays
- [x] arrow functions
- [x] function parameters
- [x] `forEach`
- [x] nested for loops
- [x] if/else statements
- [x] `.length` property
- [x] const and let 
- [x] nested if statements
- [x] `++`
- [x] `===`
- [x] assignment operator
- [x] `reduce` method
- [x] object literals
- [x] strict inequality operator
- [x] class syntax
- [x] constructor
- [x] this keyword
- [x] booleans
- [x] `textContent`
- [x] `classList`
- [x] add method
- [x] toggle method
- [x] `e.target`
- [x] `addEventListener`
- [x] contains method
- [x] Math.ceil
- [x] Math.random
- [x] createElement
- [x] `Object.entries`
- [x] array destructuring 
- [x] toUpperCase
- [x] slice
- [x] replace
- [x] createTextNode
- [x] template literals
- [x] appendChild
- [x] Object.values
- [x] break keyword
- [x] Object.keys
- [x] remove method
- [x] alert
- [x] confirm
- [x] reload method
- [x] new keyword
- [x] setTimeout

Checklist:


scissorsneedfoodtoo:
Hey @jdwilkin4, thanks for your patience. This is looking great, and the UI is much more in-line with the rest of the projects now.

Just noticed a few things and will leave some review comments.

intermediate-algorithm-scripting-yahtzee-game/index.html
```diff
@@ -1,104 +1,126 @@
...
-      simpler:
-    </p>
-
-    <ul>
-      <li>There will be 6 rounds or turns, each comprised of three dice rolls</li>
-      <li>Score combinations will be only from the lower section</li>
-      <li>No bonus scores of any kind</li>
-    </ul>
+    <main>
+      <div id="game">
```
scissorsneedfoodtoo, line 16: Since the game has a few different inputs and buttons, it might be a good idea to turn it into a form. I think the easiest way would be to change this `div` into a `form` element:

```suggestion
      <form id="game">
```

intermediate-algorithm-scripting-yahtzee-game/index.html
```diff
@@ -1,104 +1,126 @@
...
-        <div class="die selected"></div>
-        <div class="die selected"></div>
-        <div class="die selected"></div>
-        <div class="die selected"></div>
-        <div class="die selected"></div>
+        <button class="btn" id="keep-score-btn" disabled>
+          Keep the above selected score
+        </button>
+        <button class="btn" id="roll-dice-btn">Roll the selected dice</button>
       </div>
```
scissorsneedfoodtoo, line 117: Closing tag for the `div` to `form` suggestion above.

```suggestion
      </form>
```

intermediate-algorithm-scripting-yahtzee-game/index.html
```diff
@@ -1,104 +1,126 @@
...
-      <li>No bonus scores of any kind</li>
-    </ul>
+    <main>
+      <div id="game">
+        <div id="dice">
+          <button class="die selected"></button>
+          <button class="die selected"></button>
+          <button class="die selected"></button>
+          <button class="die selected"></button>
+          <button class="die selected"></button>
```
scissorsneedfoodtoo, line 22: If we decide to go with an HTML form, then I believe these buttons will all default to submit buttons. One way to prevent the from from submitting whenever a button is with the `type` attribute:

```suggestion
          <button class="die selected" type="button"></button>
          <button class="die selected" type="button"></button>
          <button class="die selected" type="button"></button>
          <button class="die selected" type="button"></button>
          <button class="die selected" type="button"></button>
```

intermediate-algorithm-scripting-yahtzee-game/index.html
```diff
@@ -1,104 +1,126 @@
...
-      <div id="dice">
-        <div class="die selected"></div>
-        <div class="die selected"></div>
-        <div class="die selected"></div>
-        <div class="die selected"></div>
-        <div class="die selected"></div>
+        <button class="btn" id="keep-score-btn" disabled>
+          Keep the above selected score
+        </button>
+        <button class="btn" id="roll-dice-btn">Roll the selected dice</button>
```
scissorsneedfoodtoo, line 116: Same suggestion as above to set the `type` attribute:

```suggestion
        <button class="btn" id="keep-score-btn" type="button" disabled>
          Keep the above selected score
        </button>
        <button class="btn" id="roll-dice-btn" type="button">Roll the selected dice</button>
```

intermediate-algorithm-scripting-yahtzee-game/index.html
```diff
@@ -1,104 +1,126 @@
...
-      select / unselect a die by clicking on it. <br> On the first roll of every round you need to roll all the five
-      dice. Now, let's get rolling!</p>
-  </header>
+        <p class="rounds-text">
+          <strong>Rolls:</strong> <span id="current-round-rolls">0</span> |
+          <strong>Round:</strong> <span id="current-round">1</span>
+        </p>
 
-  <hr>
+        <div id="score-options">
```
scissorsneedfoodtoo, line 30: At first I was thinking that this might also be better as a `fieldset` element to group these radio inputs, but it's probably fine the way it is. The fact that all these elements are part of the same form should be enough to group them together.

intermediate-algorithm-scripting-yahtzee-game/index.js
```diff
@@ -1,9 +1,19 @@
...
+const diceButtons = document.querySelectorAll("#dice > button");
+const diceContainer = document.getElementById("dice");
+const currentRoundText = document.getElementById("current-round");
+const currentRoundRollsText = document.getElementById("current-round-rolls");
+const totalScoreText = document.getElementById("total-score");
+const scoreHistory = document.getElementById("score-history");
+const rollDiceBtn = document.getElementById("roll-dice-btn");
+const keepScoreBtn = document.getElementById("keep-score-btn");
+
+// Good review of selection Sort algorithm
```
scissorsneedfoodtoo, line 13: Agreed, it would be good to review selection sort here. Also, it's good that it introduces the concept of creating deep copies of incoming arrays.

intermediate-algorithm-scripting-yahtzee-game/index.js
```diff
@@ -25,36 +36,32 @@ const selectionSort = (arr) => {
...
 
-// Sum the elements of an integer array
-const sumOfArrElements = (arr) => {
-  let arrElementsSum = 0;
-  for (let i = 0; i < arr.length; i++) {
-    arrElementsSum += arr[i];
-  }
-  return arrElementsSum;
-};
+const sumOfArrElements = (arr) => arr.reduce((prev, curr) => prev + curr, 0);
```
scissorsneedfoodtoo, line 64: This looks like great extra practice with `reduce()`.

scissorsneedfoodtoo:
I just took a look at the JS and everything LGTM. There's some good review there, and a lot of algorithmic thinking involved in building a functional Yahtzee game. And the way you simplified a few things looks great.

One last thing I noticed is that the game inputs overflow from the container on smaller screen sizes. I think it has something to do with the CSS Grid layout applied to the `main` element. Commenting those out seems to fix the layout, though it breaks some other things like the spacing of the dice and the margin of the scores `div`.

RandellDawson:
The game does not seem to play as I expect.  If I select a dice, it turns white.  Then if I click `Roll the selected dice`, all the other dice not selected changed instead of the one I selected. Or is the idea that green background means a dice is already selected?  If so, that should probably be explained.

jdwilkin4:
@RandellDawson 
The green die are the selected ones. 
When you click on it then it is no longer selected.
The original version, as you can see in the sandbox listed the rules for this simplified version of the game.
https://lihjgs.csb.app/

The plan was to show them a list of rules in the first step instead of having it always displayed in the app itself.
To me, it looks to text heavy to have the rules constantly showing like this, so that is why I omitted.

intermediate-algorithm-scripting-yahtzee-game/index.html
jdwilkin4: I personally like keeping it as a div because it provides more flexibility for styling options.

jdwilkin4:
@scissorsneedfoodtoo 

Thank you for reviewing the project.
I'll take another look at the grid layout so I can try to resolve the responsive issues for smaller devices

naomi-lgbt:
> The plan was to show them a list of rules in the first step instead of having it always displayed in the app itself.

What if we list the rules in the app but make them togglable? We could use a button to display them with JS?

jdwilkin4:
@naomi-lgbt 

Would we have the campers build that or just supply it to them?

naomi-lgbt:
> Would we have the campers build that or just supply it to them?

I'd say we would supply the HTML/CSS, and have them build the JS. This is the fourth project in the list, so it's likely here that we'd stop having campers write the HTML and CSS? 🤔

jdwilkin4:
@naomi-lgbt 

Yeah, I agree with supplying the HTML and CSS for them.
I just wasn't sure if we wanted the functionality for the instructions menu already built for them for or have them built it out as part of the first few steps of the project.

It would be a good learning opportunity for campers

naomi-lgbt:
> I just wasn't sure if we wanted the functionality for the instructions menu already built for them or have them built it out as part of the first few steps of the project.

Yeah, I'd say we have them build the functionality.

> "The rules text is rather large - you can give your users a way to hide it with a button. The HTML and CSS are prepared for you, but you'll need to write the JavaScript to make the button work. Start by declaring a `toggleRules` function."

Or some such.

bbsmooth:
I'll second @RandellDawson's comments, it is not intuitive which dice are selected and which aren't. Dice are usually white, so I think it makes sense to have them always be white and the default state be NOT selected. Then the user can select which ones they want to roll again by clicking on them, and you could add something like a bright outline and/or color change. To me this mimics the real game much better.

jdwilkin4:
I can look into reworking the dice logic and reversing it from the original app.

>Also, would you mind some accessibility feedback? Kind of like what I gave for the music player. Not a complete audit but just pointing out some of the big stuff.

You can leave a review. Thanks 👍 

But also feel free to leave reviews on any open PR or future ones posted in this repo.
Now is the perfect time for anyone to leave comments on these prototypes so they can get into a stable state to be broken down into steps. 👍

jdwilkin4:
Quincy brought up a good point about naming this Yahtzee game and it interfering with possible trademark issues.
So we will rename this project as a dice game. 
We already have a dice game in the curriculum, so we can rename this advanced dice game.

scissorsneedfoodtoo (approved):
Thanks again for your patience, @jdwilkin4.

Took a look through all of your updates and everything LGTM. I think this is a good direction for this project to take. I like the way you simplified the rules so that the game is similar enough to Yahtzee to teach the concepts covered here, but different enough so that Hazbro's lawyers won't come knocking :)

Also, even though there are some different buttons here, I don't think there are any accessibility issues. I was able to tab through the page and use my keyboard to select regular and radio buttons, and play a few rounds.

I noticed a couple of minor things and will leave them as review suggestions. Otherwise I think this is ready to merge.

intermediate-algorithm-scripting-advanced-dice-game/index.js
```diff
@@ -84,54 +93,37 @@ class Game {
...
     this.rollsInCurrentRound++;
   }
 
   updateStatsUI() {
-    document.querySelector("#current-round").textContent = this.currentRound;
-    document.querySelector("#total-score").textContent = this.totalScore;
-    document.querySelector(
-      "#current-round-rolls"
-    ).textContent = this.rollsInCurrentRound;
+    currentRoundText.textContent = this.currentRound;
```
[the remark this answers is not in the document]
naomi-lgbt: This would be a good "error step", tbh - have them leave the `if` statement off, point out the error, teach them how to fix.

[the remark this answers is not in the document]
scissorsneedfoodtoo: Awesome, that's a good idea @naomi-lgbt.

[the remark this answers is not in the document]
jdwilkin4: @scissorsneedfoodtoo 

I added a comment in the codebase to add an error step 👍

jdwilkin4:
@scissorsneedfoodtoo 

I have added in the suggestions 👍

PR #322, opened 2022-08-11
fix: accessiblity issues for music player

### LINKED ISSUE #319: music-playlist-app accessibility enhancements
Hey @jdwilkin4, I'm starting with the big stuff first. I know that accessibility probably isn't the main concern for this project. But it is also nice to set a good example.

- [x] The four control buttons at the top need text associated with them. I would probably use sr-only text but you can use aria-label if you prefer. Additionally, the text for the play/pause button should be updated depending on the status of the player. I would also recommend that the text for the play button include the name of the song to be played. For example, when the app first loads, the play button will play the first song in the list, so the sr-only text should say "play [song name]" with [song name] being the first song in the list.
- [x] The playlist itself feels like it should be coded as an HTML list. This would also automatically tell screen reader users how many songs are in the list. There is another advantage to using a list here which I won't go into now.
- [x] The delete button for each song should have sr-only text that includes the name of the song to be deleted. That way screen readers will hear "Delete [song name]". This is especially important if they are navigating by Tab or listing all buttons.
- [x] Adding a click handler to a div is inaccessible, especially for keyboard users. Any element that can be clicked must have a well-defined role and accessible name. The better option here is to probably use a button, since this does seem to be acting as a play button for the song. I wouldn't have two `p` elements in there, just text, perhaps something like "Play [song name] by [artist]". The words "Play" and "by" would be sr-only. Add `span`s where needed for styling.
- [x] When you mouse over a song row, the entire row changes color but only clicking over the song name/artist text triggers the song to play. In other words, there are gray areas around the delete button that do nothing. There are various options for handling this, I'll let you decide on which one you like best.

You might not consider all of these to be important, I completely understand. But I would definitely implement the sr-only text for the four buttons and removing the click handler on the `div`s.

jdwilkin4:
Checklist:

Closes #319


scissorsneedfoodtoo (commented):
Thanks again for your patience @jdwilkin4, and for all of your helpful details and suggestions in #319, @bbsmooth.

Tested this locally and everything looks great. It seems like all of the points in #319 were addressed. `aria-label` attributes are present and updated for the control buttons at the top, and the buttons for the songs and the delete button for each all have `sr-only` text. I also tabbed around and everything seems to be keyboard accessible now.

The only thing I noticed is that the play / pause button at the top begins with its `aria-label` attribute set to "play and pause song button". I think that makes sense since people could shuffle the songs before pressing that play button, but if there's a better approach, please let us know @bbsmooth.

music-playlist-app/index.html
```diff
@@ -96,7 +116,88 @@ <h1>Music Playlist Project</h1>
...
+                  type="button"
+                  id="we-are-going-to-make-it"
+                  class="song-btn"
+                >
+                  <span class="sr-only">Play</span> We Are Going to Make it
+                  <span class="sr-only">by</span>
+                  - Quincy Larson
+                </button>
+                <div class="btn-container">
+                  <span class="sr-only">Delete We Are Going to Make it</span>
```
bbsmooth, line 132:
```suggestion
                  <span class="sr-only">Delete We Are Going to Make it</span>
```

This `span` needs to be inside the `button` text so that the screen reader will announce "Delete We Are Going to Make it" when the user focuses the button. Same for the other four Delete buttons.

music-playlist-app/index.html
```diff
@@ -75,7 +90,12 @@ <h1>Music Playlist Project</h1>
                 <line x1="19" y1="5" x2="19" y2="19"></line>
               </svg>
             </button>
-            <button id="shuffle-btn" class="btn-styles">
+            <button
+              type="button"
+              aria-label="shuffle songs button"
```
bbsmooth, line 95:
```suggestion
              aria-label="shuffle songs"
```

music-playlist-app/index.html
```diff
@@ -61,7 +71,12 @@ <h1>Music Playlist Project</h1>
                 </g>
               </svg>
             </button>
-            <button id="next-btn" class="btn-styles">
+            <button
+              type="button"
+              aria-label="next song button"
```
bbsmooth, line 76:
```suggestion
              aria-label="next song"
```

music-playlist-app/index.html
```diff
@@ -21,7 +21,12 @@ <h1>Music Playlist Project</h1>
             <div class="text-container"></div>
           </div>
           <div class="media-controls">
-            <button id="previous-btn" class="btn-styles">
+            <button
+              type="button"
+              aria-label="previous song button"
```
bbsmooth, line 26:
```suggestion
              aria-label="previous song"
```

bbsmooth:
The next song button rolls over to the beginning of the list after you reach the last song but the previous song button does not roll over to the end of the list after you reach the first song.

bbsmooth:
I would recommend we remove all of the hard coded list items in the HTML and just call `createPlaylist` in the JS to populate the playlist when the page loads.

music-playlist-app/index.js
```diff
@@ -83,8 +86,19 @@ function displaySong(song = "", artist = "") {
 }
 
 function playPauseSong() {
-  isPlaying ? song.pause() : song.play();
-
+  if (isPlaying) {
+    song.pause();
+    playPauseBtn.setAttribute(
+      "aria-label",
+      `Paused ${playlist[currentSong].title}`
```
bbsmooth, line 93:
```suggestion
      `Play ${playlist[currentSong].title}`
```
I know this seems a little counter intuitive, but the button needs to describe what it will do, not the state of the player. So when the song is paused then the button will play the song and thus the accessible name should be "Play [song title]".

music-playlist-app/index.js
```diff
@@ -83,8 +86,19 @@ function displaySong(song = "", artist = "") {
...
+    );
+  } else {
+    song.play();
+    playPauseBtn.setAttribute(
+      "aria-label",
+      `Playing ${playlist[currentSong].title}`
```
bbsmooth, line 99:
```suggestion
      `Pause ${playlist[currentSong].title}`
```

Same logic as above, when the song is playing then the button should read "Pause [song name]" since that's what will happen when you click on the button.

music-playlist-app/index.js
```diff
@@ -103,6 +121,10 @@ function goToNextSong() {
   currentSong === playlist.length - 1 ? (currentSong = 0) : currentSong++;
   song = new Audio(playlist[currentSong].audio);
   song.play();
+  playPauseBtn.setAttribute(
+    "aria-label",
+    `Playing ${playlist[currentSong].title}`
```
bbsmooth, line 126:
```suggestion
    `Pause ${playlist[currentSong].title}`
```

Again, the button should say what it is going to do, not the state of the player.

music-playlist-app/index.js
```diff
@@ -95,6 +109,10 @@ function goToPreviousSong() {
   currentSong === 0 ? (song.currentTime = 0) : currentSong--;
   song = new Audio(playlist[currentSong].audio);
   song.play();
+  playPauseBtn.setAttribute(
+    "aria-label",
+    `Playing ${playlist[currentSong].title}`
```
bbsmooth, line 114:
```suggestion
    `Pause ${playlist[currentSong].title}`
```

bbsmooth:
At the end of the `shufflePlaylist` function I would add the following to update the play/pause button to reflect the new first song:

```
playPauseBtn.setAttribute(
    "aria-label",
    `Play ${playlist[currentSong].title}`
  );
```

bbsmooth:
My thinking on the play/pause button is that it always reflects the current song to be played/paused. So when the page first loads the play button's accessible name (i.e. the `aria-label`) should be "Play [first song name in list]". And when the user shuffles the songs then the play button should be updated to point to the song at the top of the list (which my code suggestion above will do).

So the issue here is to make sure that on initial page load the play button is set correctly.

bbsmooth:
We might think about splitting the play/pause button icon into separate icons and then show the appropriate icon based on the status of the button (i.e. show the play icon when the player is paused and the pause icon when the player is playing a song). Granted, most people are going to hear that the music is playing and so they will know what the button is going to do when they click it. But that is not a 100% guarantee, so having the button only show one icon will make this clear for people who can't hear.

jdwilkin4:
@bbsmooth 

That you for your review.

I have started to implement the changes and will continue to implement the rest after work.

jdwilkin4:
@bbsmooth 

All of the points you mentioned have been implemented.

scissorsneedfoodtoo:
Hey @jdwilkin4, thanks again for your patience. And thank you @bbsmooth for your thorough review. I had no idea that buttons for media apps should behave that way. Really great information you shared with us there.

I noticed a few things while reviewing that I'll leave as review comments.

music-playlist-app/index.js
```diff
@@ -83,8 +90,19 @@ function displaySong(song = "", artist = "") {
 }
 
 function playPauseSong() {
-  isPlaying ? song.pause() : song.play();
-
+  if (isPlaying) {
```
scissorsneedfoodtoo, line 93: The last thing I noticed is that, even after deleting all songs, it's still possible to play a song.

What might be good is to do an early return if the length of `playlist` is 0:

```suggestion
  if (playlist.length === 0) return;
  
  if (isPlaying) {
```

Early returns could also be added to `goToNextSong` and `goToPreviousSong`, which play a song after changing tracks.

music-playlist-app/index.js
```diff
@@ -128,6 +154,11 @@ function shufflePlaylist() {
...
   deleteSong();
   chooseSong();
+
+  playPauseBtn.setAttribute(
+    "aria-label",
+    `Play ${playlist[currentSong].title}`
+  );
 }
 
 function deleteSong() {
```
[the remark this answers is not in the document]
jdwilkin4: Replaced `splice` with using filter instead.
I tested it locally and it should be fixed now

bbsmooth (approved):
The four accessibility issues in [#319](https://github.com/freeCodeCamp/CurriculumExpansion/issues/319) are satisfied now!

bbsmooth:
@jdwilkin4 I am finding an issue with the delete button though as I was testing accessibility. If you delete the first two songs in the list and then try to play the third song (which is now the first song in the list) it won't play and you'll get the following console error:

```
Uncaught TypeError: playlist[currentSong] is undefined
```

jdwilkin4:
@bbsmooth 
Thank you for catching this. 

I have created a new issue to fix this bug and do some more testing on this app to find any more bugs. 
https://github.com/freeCodeCamp/CurriculumExpansion/issues/327

I am going to go ahead and merge this since issues brought up in #319 have been addressed.

PR #324, opened 2022-09-06
feat: updating regex spam filter project

jdwilkin4:
I updated this spam regex project that was in the repo. 
I added some styles and added all of the javascript into one file. 

Live app:
https://zpb344.csb.app/
Checklist:

Concepts covered

- [x] `getElementById`
- [x] const
- [x] arrays
- [x] regex character groups
- [x] regex i flag
- [x] regex g flag
- [x] match method
- [x] test method
- [x] `\s`
- [x] non capturing group
- [x] some method
- [x] arrow functions
- [x] `addEventListener`
- [x] `textContent`
- [x] `alert`
- [x] ternary operator
- [x] booleans
- [x] strict equality
- [x] strings


learn-regular-expressions-by-building-a-spam-filter/index.js
```diff
@@ -0,0 +1,34 @@
...
+
+const blacklistRegexps = [
+  /[0-9] hundred|thousand|million|billion dollars/gi,
+  /PLEASE HELP|ASSIST ME/i,
+  /(?:^|\s)fr[e3][e3] m[o0]n[e3]y(?:$|\s)/i,
+  /(?:^|\s)[s5][t7][o0][c{[(]k [a@4]l[e3]r[t7](?:$|\s)/i,
+  /(?:^|\s)d[e3][a@4]r fr[i1|][e3]nd(?:$|\s)/i,
+];
+
+const isSpam = (msg) => blacklistRegexps.some((re) => re.test(msg));
```
naomi-lgbt, line 22: We could add a step that has the camper use `match` here first, then explain that because we don't need to do anything with the matched value, moving to `test` could be more performant.

jdwilkin4:
TODO: 

Need to add example with look aheads and look behinds

learn-regular-expressions-by-building-a-spam-filter/index.js
```diff
@@ -0,0 +1,40 @@
...
+ * I read through the previous discussions for this project
+ * and we can start with matches for literal phrases and then have the
+ * campers create more complex regex which seemed to be the original intent for this project.
+ *
+ *
+ *
+ */
+
+const blacklistRegexps = [
+  /[0-9] hundred|thousand|million|billion dollars/gi,
```
scissorsneedfoodtoo, line 15: Currently this works well. It matches "thousand" in "100 thousand dollars".

If we want to show how to match everything in "100 million dollars",  we could build this line up to something like `\d*\s+(hundred|thousand|million|billion)(\s+dollars)?`.

It could start off exactly like it is now, then get a bit more complex by adding a capturing groups and using character classes like `\d` to replace `[0-9]`, and special flags like `*` and `?`.

`\s` is probably not necessary, and can be replaced with a whitespace character. I just find `\s` a bit easier to read.

learn-regular-expressions-by-building-a-spam-filter/index.js
```diff
@@ -0,0 +1,40 @@
...
+  /PLEASE HELP|ASSIST ME/i,
```
scissorsneedfoodtoo, line 16: This would be a great expression to start with.

If we want to teach how to match a whole phrase like "please help me", this would be a good place to introduce a capture group around `HELP|ASSIST`.

scissorsneedfoodtoo:
> TODO:
> 
> Need to add example with look aheads and look behinds

It would be cool to teach lookaheads if possible.

For positive lookaheads, something like `/\d+(?=\s+dollars)/gi` would match "100" in "100 dollars", and "000" in "1,000 dollars". We could also look for other currencies like euros that often come up in spam emails with `/\d+(?=\s+(dollars|euros))/gi`. We could even search for USD or EUR.

I haven't been able to think of a great way to teach negative lookaheads. One way might be to search for some currencies that are unlikely to appear in spam emails, like `/\d{3}(?!\s(KRW|JPY))/gi`, but unfortunately that captures all other groups of 3 digits that aren't followed by a space then KRW or JPY.

I don't think we should teach lookbehinds, though. Unfortunately Safari still doesn't support them.

learn-regular-expressions-by-building-a-spam-filter/index.js
```diff
@@ -0,0 +1,40 @@
...
+/**
+ * I read through the previous discussions for this project
+ * and we can start with matches for literal phrases and then have the
+ * campers create more complex regex which seemed to be the original intent for this project.
+ *
+ *
+ *
+ */
+
+const blacklistRegexps = [
```
naomi-lgbt, line 14: Given the discourse around `whitelist`/`blacklist`, do we want to change this to `denyList`?

scissorsneedfoodtoo: Thanks for catching this @naomi-lgbt.

I agree. `denyList`, `blockList`, or something like `denyListRegexps` would be a better variable name.

learn-regular-expressions-by-building-a-spam-filter/index.js
jdwilkin4: We could add some steps that refactors this ` /[0-9] hundred|thousand|million|billion dollars/gi` to slowly introduce the `\d`, `*` and `?` flags. 
I can add a comment in the codebase about this.

learn-regular-expressions-by-building-a-spam-filter/index.js
jdwilkin4: I like denyListRegexps 👍

learn-regular-expressions-by-building-a-spam-filter/index.js
jdwilkin4: I can add a comment in the codebase about adding a step to refactor this to include the capturing group

jdwilkin4:
I have updated the codebase with comments about possible refactoring and additional steps we could include for campers.

scissorsneedfoodtoo (approved):
Thanks again for your patience @jdwilkin4.

Took another look at your project and everything LGTM. This is one of those projects where most of the instruction will be in the refactoring, so thank you for adding all of your detailed notes, and for your helpful contents covered checklist above.

PR #334, opened 2023-02-06
feat: add Piet Mondriaan intermediate CSS project

Sboonny:
Checklist:

### DIFF
--- a/Intermediate-css-piet-mondriaan/css/styles.css
+++ b/Intermediate-css-piet-mondriaan/css/styles.css
@@ -1,169 +1,169 @@
 .grid {
-    display: grid;
-  }
-  
-  .grid-stack-area {
-    grid-template-areas: 'stack';
-  }
-  
-  .grid-area-name {
-   grid-area: stack;
-  }
-  
-  .silcing-raw-big {
-    grid-template-rows: .4fr repeat(4, 1fr) .75fr;
-  }
-  
-  .silcing-columns-big {
-    grid-template-columns: 0.2fr 1.2fr repeat(2, 1fr);
-  }
-  
-  .margin-0 {
-    margin: 0
-  }
-  
-  .margin-top-2 {
-    margin-top: 0.5em;
-  }
-  
-  .margin-right-2 {
-    margin-right: .5em
-  }
-  
-  .margin-bottom-2 {
-    margin-bottom: .5em
-  }
-  
-  .border-top-width {
-    border-top-width: 1em;
-  }
-  
-  .border-right-width {
-    border-right-width: 1em;
-  }
-  
-  .border-bottom-width {
-    border-bottom-width: 1em;
-  }
-  
-  .border-left-width {
-    border-left-width: 1em;
-  }
-  
-  .border-top-style-solid {
-    border-top-style: solid;
-  }
-  
-  .border-right-style-solid {
-    border-right-style: solid;
-  }
-  
-  .border-bottom-style-solid {
-    border-bottom-style: solid;
-  }
-  
-  .border-left-style-solid {
-    border-left-style: solid;
-  }
-  
-  .border-color-yellow {
-    border-color: hsl(59, 100%, 50%);
-  }
-  
-  .border-color-brown {
-    border-color: hsl(7, 100%, 50%);
-  }
-  
-  .border-color-blue {
-    border-color: hsl(240, 63%, 62%);
-  } 
-  
-  .border-top-color-brown {
-    border-top-color: hsl(7, 100%, 50%);
-  }
-  
-  .height-2 {
-    height: 20%;
-  }
-  
-  .height-6 {
-    height: 60%;
-  }
-  
-  .height-8 {
-    height: 80%;
-  } 
-  
-  .height-10 {
-    height: 100%;
-  }
-  
-  .height-full {
-    height: 100vh;
-  }
-  
-  .transform-top {
-    transform: translate(0, -5%);
-  }
-  
-  .transform-bottom {
-    transform: translate(0, 50%);
-  }
-  
-  .position-relative {
-    position: relative;
-  }
-  
-  .position-absolute {
-    position: absolute
-  }
-  
-  .right-low-5 {
-    right: 5%
-  }
-  
-  .right-2 {
-    right: 20%
-  }
-  
-  .right-3 {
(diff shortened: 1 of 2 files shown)


scissorsneedfoodtoo:
Hey @Sboonny, thanks for your patience.

Like we mentioned in chat, this project looks great, and is quite a faithful reproduction of the original painting.

I took a look and the main issue is that the code currently uses CSS Grid to place the different lines.

Looking at the [project-based curriculum spreadsheet](https://docs.google.com/spreadsheets/d/1KXEJvgnAziiePMp2Ovf_ssDqCFFWYuycXWBNp91k3qM), this project should ideally cover as many of the concepts that the current Picasso project does (column Q), and not introduce concepts that are covered in future projects.

The things that stand out to me are that CSS Grid is introduced in the CSS Grid Magazine project a little later. Also, the Picasso project this is replacing is the first time that Font Awesome and the `i` element are introduced.

I don't think it's necessary to introduce Font Awesome and the `i` element here . If you can think of a way to add in an FA icon that feels natural, we can go with that. Otherwise, I think it's fine to move the introductions of those things to the CSS Grid project where they're currently reviewed.

Whatever the case, we want to make as few changes as possible to other projects in the RWD cert.

I played around with some your code and came up with a simple example of a few lines from the painting in a small frame that's centered on the screen. I did that so the whole painting is visible in the preview pane on desktop screens, but we could also use the full width and height of the view like you are currently.

I'll leave the HTML and CSS as a barebones example that might help with refactoring later.

Intermediate-css-piet-mondriaan/css/styles.css
```diff
@@ -0,0 +1,169 @@
...
+
+a::before {
+  content: "";
+  width: 100%;
+  height: 100%;
+  position: absolute;
+  cursor: pointer;
+  inset: 0;
+  z-index: 999;
+}
```
scissorsneedfoodtoo, line 169: There are probably better names for these classes, especially `height`, but it might be a good jumping off point for later:

```css
*,
*::before,
*::after {
  box-sizing: border-box;
}

* {
  margin: 0;
}

html,
body,
main {
  height: 100%;
}

main {
  display: flex;
  justify-content: center;
  align-items: center;
}

.frame {
  height: 540px;
  width: 490px;
  border: 20px solid black;
  position: relative;
}

.vertical,
.horizontal {
  position: absolute;
}

.vertical {
  width: 15px;
  height: 100%;
}

.horizontal {
  height: 15px;
  width: 100%;
}

.yellow {
  background-color: hsl(59, 100%, 50%);
}

.red {
  background-color: hsl(7, 100%, 50%);
}

.blue {
  background-color: hsl(240, 63%, 62%);
}

.height-1 {
  z-index: 10;
}

.height-2 {
  z-index: 20;
}

.height-3 {
  z-index: 30;
}

.height-4 {
  z-index: 40;
}

.height-5 {
  z-index: 50;
}

.height-6 {
  z-index: 60;
}

.height-7 {
  z-index: 70;
}

.height-8 {
  z-index: 30;
}

.line-a {
  top: 2%;
}

.line-b {
  left: 2%;
}

.line-c {
  left: 8%;
}
```

Intermediate-css-piet-mondriaan/index.html
```diff
@@ -0,0 +1,76 @@
...
+    </div>
+    <p class="grid-area-name transparent">This page is copying the work of Piet Mondrain, check out <a
+        class="transparent" href="https://www.pietmondrian.eu/images/imgs-480-eng/new-york-city-1942-mondrian.jpg"
+        target="_blank">his
+        wikipedia</a></p>
+  </main>
+
+</body>
+
+</html>
```
scissorsneedfoodtoo, line 76:
```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="utf-8" />
    <link rel="stylesheet" href="./styles.css" />
    <title>Piet Mondriaan – New York City, 1942</title>
  </head>
  <body>
    <main>
      <div class="frame">
        <div class="line-a height-8 horizontal red"></div>
        <div class="line-b height-7 vertical yellow"></div>
        <div class="line-c height-1 vertical blue"></div>
      </div>
      <!-- <p class="grid-area-name transparent">
        This page is copying the work of Piet Mondrain, check out
        <a
          class="transparent"
          href="https://www.pietmondrian.eu/images/imgs-480-eng/new-york-city-1942-mondrian.jpg"
          target="_blank"
          >his wikipedia</a
        >
      </p> -->
    </main>
  </body>
</html>
```

scissorsneedfoodtoo (commented):
Hey @Sboonny, thanks for your patience.

Like we mentioned in chat, this project looks great, and is quite a faithful reproduction of the original painting.

I took a look and the main issue is that the code currently uses CSS Grid to place the different lines.

Looking at the [project-based curriculum spreadsheet](https://docs.google.com/spreadsheets/d/1KXEJvgnAziiePMp2Ovf_ssDqCFFWYuycXWBNp91k3qM), this project should ideally cover as many of the concepts that the current Picasso project does (column Q), and not introduce concepts that are covered in future projects.

The things that stand out to me are that CSS Grid is introduced in the CSS Grid Magazine project a little later. Also, the Picasso project this is replacing is the first time that Font Awesome and the `i` element are introduced.

I don't think it's necessary to introduce Font Awesome and the `i` element here . If you can think of a way to add in an FA icon that feels natural, we can go with that. Otherwise, I think it's fine to move the introductions of those things to the CSS Grid project where they're currently reviewed.

naomi-lgbt:
Just leaving a quick note from our curriculum meeting - we are considering something a little different like painting a cat, so it's more engaging than placing divs (we can tie in borders  and border radii to create shapes without transform)

QuincyLarson:
To confirm, I think we should replace the Picasso project with 2 projects: the Mondrian project, and the cat project. This way we can cover everything, and we can give learners additional practice at applying these concepts.

naomi-lgbt:
The only new concepts in this project are positioning elements, I don't see a need for a full second project.

Ksound22:
@naomi-lgbt do you think the cat painting would be enough to replace the Picasso painting? A bunch of what we covered in the Picasso painting project is there: https://codepen.io/vcmg/pen/OVYpWN

naomi-lgbt:
I would think so, yes.

PR #338, opened 2023-04-20
learn-basic-string-and-array-methods-by-building-a-music-player-app

RafaelDavisH:
Checklist:

- [X] I have read [freeCodeCamp's contribution 
guidelines](https://contribute.freecodecamp.org).
- [X] My pull request has a [descriptive 
title](https://contribute.freecodecamp.org/#/how-to-open-a-pull-request?id=prepare-a-good-pr-title) (**not** a vague title like `Update index.md`)

This the first draft of the music player project. It needs more string and array methods.

String and Array methods implemented

- [x] arr.find()
- [x] arr.forEach()
- [x] arr.indexOf()
- [x] arr.join()
- [x] arr.length
- [x] arr.map()
- [x] arr.sort()
- [ ] arr.splice()
- [ ] arr.push()

other concepts

- [x] audio.pause()
- [x] audio.play()
- [x] new Audio()

Bugs to fix:

- [ ] After pressing the shuffle button, the ArrowUp, ArrowDown, and Space keystrokes don't
work well. Specially, if after pressing the shuffle btn, when pressing Space it won't pause
the song and it shuffles the playlist.
- [ ] When closing the Playlist container when pressing the "X" on the UP-RIGHT corner, the
top conatainer collapse down.


RafaelDavisH:
Screenshot 2023-04-19 at 11 34 08 PM

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/styles.css
```diff
@@ -1,128 +1,244 @@
...
-  padding: 5px;
-  font-size: 1rem;
-  background-color: var(--gray100);
+button {
+  background: transparent;
   border: none;
+  color: var(--white-color, white);
+  cursor: pointer;
+  font-size: 14px;
+  outline: none;
```
bbsmooth, line 124: Disabling the `outline` is considered an accessibility anti-pattern, especially if you don't replace it with `:focus-visible`. I understand that this course isn't about teaching accessibility, but I don't think we should encourage this as it is already far too commonly misused. Unless we want to get into the specifics of styling keyboard focus outlines I would suggest we remove this.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
```diff
@@ -2,124 +2,80 @@
...
+            </div>
+          </div>
+        </div>
+      </div>
+      <div class="playlist">
+        <div class="playlist__bar">
+          <div>
+            <img src="assets/decor-lines.svg" alt="decoration lines left" />
+          </div>
+          <h1 class="playlist-title">Playlist</h1>
```
bbsmooth, line 69:
```suggestion
          <h2 class="playlist-title">Playlist</h2>
```

Generally there should only be one `h1` on a page. I think an `h2` makes sense here since the playlist is a subset of the player.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
```diff
@@ -2,124 +2,80 @@
...
-      <div class="music-player">
-        <div class="flex-container">
-          <div class="current-song-container">
-            <img class="img-styles" alt="freeCodeCamp logo" src="./mic.jpg" />
-            <div class="text-container"></div>
+    <div class="container">
+      <div class="player">
+        <div class="player__bar">
+          <div>
+            <img src="assets/decor-lines.svg" alt="decoration lines left" />
```
bbsmooth, line 22:
```suggestion
            <img src="assets/decor-lines.svg" alt="" />
```

These are purely decorative images and should be hidden from screen reader users by using empty `alt` text.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
```diff
@@ -2,124 +2,80 @@
...
-                  />
-                </g>
-              </svg>
-            </button>
+          <h1 class="fcc-title">freeCodeCamp</h1>
+          <div>
+            <img src="assets/decor-lines.svg" alt="decoration lines right" />
```
bbsmooth, line 26:
```suggestion
            <img src="assets/decor-lines.svg" alt="" />
```

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
```diff
@@ -2,124 +2,80 @@
...
+                <img src="assets/loop.svg" alt="repeat button" />
+              </button>
+            </div>
+          </div>
+        </div>
+      </div>
+      <div class="playlist">
+        <div class="playlist__bar">
+          <div>
+            <img src="assets/decor-lines.svg" alt="decoration lines left" />
```
bbsmooth, line 67:
```suggestion
            <img src="assets/decor-lines.svg" alt="" />
```

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
```diff
@@ -2,124 +2,80 @@
...
+          </div>
+          <h1 class="playlist-title">Playlist</h1>
+          <div>
+            <img src="assets/decor-lines.svg" alt="decoration lines right" />
```
bbsmooth, line 71:
```suggestion
            <img src="assets/decor-lines.svg" alt="" />
```

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
```diff
@@ -2,124 +2,80 @@
...
-              </svg>
-            </button>
+          <h1 class="fcc-title">freeCodeCamp</h1>
+          <div>
+            <img src="assets/decor-lines.svg" alt="decoration lines right" />
           </div>
-
-          <ul class="playlist-container"></ul>
+          <div>
+            <img src="assets/menu.svg" alt="menu button" />
```
bbsmooth, line 29: What is the purpose of this image? Is this going to be an actual button at some point, or is this just for looks?

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
```diff
@@ -2,124 +2,80 @@
...
+            />
+          </div>
+          <div class="player__display">
+            <div class="player__display--songArtist">
+              <p id="player__song-title">Cruising for a Musing</p>
+              <p id="player__song-artist">Quincy Larson</p>
+            </div>
+            <div class="player__buttons">
+              <button class="rewind">
+                <img src="assets/rewind.svg" alt="rewind button" />
```
bbsmooth, line 46:
```suggestion
                <img src="assets/rewind.svg" alt="rewind" />
```
We do not need to add "button" to the alt text because screen readers will already announce that this is a button.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
```diff
@@ -2,124 +2,80 @@
...
+              </button>
+              <button class="play">
+                <img src="assets/play.svg" alt="play button" />
```
bbsmooth, line 49:
```suggestion
                <img src="assets/play.svg" alt="play" />
```

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
```diff
@@ -2,124 +2,80 @@
...
+              </button>
+              <button class="forward">
+                <img src="assets/forward.svg" alt="forward button" />
```
bbsmooth, line 52:
```suggestion
                <img src="assets/forward.svg" alt="forward" />
```

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
```diff
@@ -2,124 +2,80 @@
...
+              </button>
+              <button class="shuffle">
+                <img src="assets/shuffle.svg" alt="shuffle button" />
```
bbsmooth, line 55:
```suggestion
                <img src="assets/shuffle.svg" alt="shuffle" />
```

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
```diff
@@ -2,124 +2,80 @@
...
+              </button>
+              <button class="repeat">
+                <img src="assets/loop.svg" alt="repeat button" />
```
bbsmooth, line 58:
```suggestion
                <img src="assets/loop.svg" alt="repeat" />
```

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
```diff
@@ -2,124 +2,80 @@
...
+      <div class="playlist">
+        <div class="playlist__bar">
+          <div>
+            <img src="assets/decor-lines.svg" alt="decoration lines left" />
+          </div>
+          <h1 class="playlist-title">Playlist</h1>
+          <div>
+            <img src="assets/decor-lines.svg" alt="decoration lines right" />
+          </div>
+          <div class="playlist__close">X</div>
```
bbsmooth, line 73: Again, I know this course isn't primarily about accessibility, but adding a click handler on a `div` is inaccessible and is a pattern that should be taught at FCC. At the least, this needs to be an actual `button` with a click handler.

To make it really accessible then it needs to have an `aria-label` attribute with a value of "Playlist" and the `aria-expanded` attribute would be set to either `true` or `false` depending on whether the playlist is currently showing.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/scripts.js
```diff
@@ -0,0 +1,234 @@
...
+updatePlayingSongBkg(currentIndex);
+
+const playPausebtn = document.querySelector(".play");
+
+function playSong(song) {
+    const oldIndex = currentIndex;
+    // using indesof method
+    const newIndex = songs.indexOf(song);
+
+    playPausebtn.innerHTML = '<img src="assets/pause.svg" alt="pause button" />'
```
bbsmooth, line 96:
```suggestion
    playPausebtn.innerHTML = '<img src="assets/pause.svg" alt="pause" />'
```

bbsmooth:
The playlist has several accessibility issues, the primary one being that you can't get to it with a keyboard. I don't think it would take much effort to increase the accessibility. I would be happy to help out with this.

Star rating systems on the other hand can definitely be made accessible but they take a lot more work. I'm wondering if it might be best to leave it out of this project as it might be too advanced and detract too much from what this course is trying to teach.

bbsmooth:
Whoops, accidentally closed this. My bad.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/styles.css
```diff
@@ -1,128 +1,244 @@
...
-  --main-background-color: #0a0a23;
-  --gray100: #ffffff;
-  --gray200: #e9e8e8;
-  --golden-yellow: #feac32;
+body {
+  background-color: var(--main-bg-color);
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 100vh;
```
bbsmooth, line 20: I would suggest we remove `height: 100vh`. Setting this makes it impossible to zoom in because the top of the player goes above the viewport and you can't scroll to get it back.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
RafaelDavisH: Yes, this will be a menu button

RafaelDavisH:
@bbsmooth, Thank you for taking the time and detail the issues that need to be change. I'll star working on these issue, totally agree, and I look forward for your input again once I am done with these changes.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/styles.css
RafaelDavisH: Agreed. I'm leaving the outline and instead, I am adding some style to it.

jdwilkin4:
Hi @RafaelDavisH ! 

I will have more time later on this week to dive into the project and leave feedback.
But for the assets folder, all of the svgs seem to be a total of 3-4 lines. So maybe we could remove the assets folder altogether and just add the svgs right into the main html file. 

Remember, by this point in the curriculum, the campers will not be adding any HTML or CSS. That will all be provided for them.

jdwilkin4:
@RafaelDavisH 

I think for the songs arr, we can just have it in the main `script.js` file instead of creating a separate file for it. 

We can have campers start to build out the array for a few steps and then provide the rest. We do something similar in the football team project.

RafaelDavisH:
> Hi @RafaelDavisH !
> 
> I will have more time later on this weekend to dive into the project and leave feedback. But for the assets folder, all of the svgs seem to be a total of 3-4 lines. So maybe we could remove the assets folder altogether and just add the svgs right into the main html file.
> 
> Remember, by this point in the curriculum, the campers will not be adding any HTML or CSS. That will all be provided for them.

Thank you @jdwilkin4 for your suggestions. I have moved the song array and deleted most of the svg images, except pause and play. These are swap when they are triggered.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/styles.css
```diff
@@ -1,128 +1,253 @@
...
+  font-family: "Roboto Mono", monospace;
+}
+
+.playlist__song-rating {
+  display: flex;
+  margin: auto;
+}
+
+.playlist__song-rate.rated {
+  fill: var(--player-highlight-color);
```
jdwilkin4, line 252: Since we don't have the star ratings anymore, we could remove these styles

jdwilkin4 (changes requested):
I think this looks good so far.
I know you are still adding more to it. 

I left some comments down below 👍

bbsmooth:
Those buttons that were converted from images to svgs don't have alt text any more so there is nothing for screen readers to announce. We have a few different options for fixing this.

1. Add `role="img"` to the svg and then add a `title` element in the svg with the text we want announced for the button. For example, the Previous button would be:
```
<button class="previous">
                <svg
                  width="24"
                  height="14"
                  viewBox="0 0 24 14"
                  fill="none"
                  xmlns="http://www.w3.org/2000/svg"
                  role="img"
                >
                  <title>Previous</title>
```
2. Use sr-only text to label the button (and of course we would need to add the sr-only ruleset to the CSS). For example, the Previous button would be:
```
<button class="previous">
                <span class="sr-only">Previous</span>
```
3. Use `aria-label` to give the button an accessible name. For example, the Previous button would be:
```
<button class="previous" aria-label="Previous">
```

Any of the above will work. I'll let others decide which would be best for the course. Personally, if I were coding this I would choose either 2 or 3. Option 1 is fine, but most screen readers will add the word "graphic" or "image" when announcing the button so you get some extra information you really don't need.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
```diff
@@ -2,124 +2,181 @@
...
+              height="10"
+              viewBox="0 0 130 10"
+              fill="none"
+              xmlns="http://www.w3.org/2000/svg"
+            >
+              <path d="M130 1H0M130 9H0" stroke="#F1BE32" stroke-width="2" />
+            </svg>
+          </div>
+          <div>
+            <button class="playlist__close">X</button>
```
bbsmooth, line 174: This button needs two thing:
- A proper accessible name, which would be "Playlist". I think the easiest way to do this would be to add `aria-label="Playlist"`.
- It needs the `aria-expanded` attribute to represent the state of the playlist. If the playlist is showing then it needs to be set to `aria-expanded="true"`. If the playlist is not showing then it needs to be set to `aria-expanded="false"`. I know this might be slightly outside the scope of this course, but this is a very important accessibility requirement and if we are going to provide a toggle mechanism for the playlist then I think we need to include this.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
```diff
@@ -2,124 +2,181 @@
...
+              height="10"
+              viewBox="0 0 130 10"
+              fill="none"
+              xmlns="http://www.w3.org/2000/svg"
             >
+              <path d="M130 1H0M130 9H0" stroke="#F1BE32" stroke-width="2" />
+            </svg>
+          </div>
+          <div>
+            <button>
```
bbsmooth, line 45: Is this button actually going to be used? Will it eventually open up a menu or is this just for looks?

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/scripts.js
```diff
@@ -0,0 +1,202 @@
...
+const songListDiv = document.getElementById("playlist__songs");
+const songPlaylistDiv = document.querySelector(".playlist");
+
+function renderSongs(songs) {
+
+// map Array Method
+// TODO: First Array for steps
+const songList = songs.map((song) => {
+    return `
+    <div class="playlist__song" role="button" tabindex="0">
```
bbsmooth, line 27:
```suggestion
    <button class="playlist__song">
```

It is best to use an actual `button` instead of adding a role and tabindex. Also, buttons are supposed to work with both the Enter key and Spacebar, so you would need to had logic to handle both of those in addition to pointer clicks. If you use an actual `button` then you can just use a click handler for everything.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/scripts.js
```diff
@@ -0,0 +1,202 @@
...
+            <p class="playlist__song-title">${song.title}</p>
+            <p class="playlist__song-artist">${song.artist}</p>
+        </div>
+        <div class="playlist__song-rating">
+
+        </div>
+        <div class="playlist__song-duration">
+            <p>${song.duration}</p>
+        </div>
+    </div>
```
bbsmooth, line 38:
```suggestion
    </button>
```

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/scripts.js
```diff
@@ -0,0 +1,202 @@
...
+})
+
+
+document.querySelectorAll('div[role="button"]').forEach((songDiv,index) => {
+    songDiv.addEventListener('keydown', event => {
+        if (event.key === 'Enter') {
+            playSelectedSong(songs[index], songDiv);
+        }
+    });
+});
```
bbsmooth, line 200:
```suggestion
document.querySelectorAll('button.playlist__song').forEach((song, index) => {
    song.addEventListener('click', event => playSelectedSong(songs[index], song));
});
```

Now that we are using actual `button`s for the songs in the list we can just use a click handler and now the Enter key, Spacebar, and any pointer clicks will start the song playing.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/scripts.js
```diff
@@ -0,0 +1,202 @@
...
+
+  // song playlist
+const songListDiv = document.getElementById("playlist__songs");
+const songPlaylistDiv = document.querySelector(".playlist");
+
+function renderSongs(songs) {
+
+// map Array Method
+// TODO: First Array for steps
+const songList = songs.map((song) => {
```
bbsmooth, line 25:
```suggestion
const songList = songs.map((song, index) => {
```

See the comment on the button below for an explanation.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/scripts.js
```diff
@@ -0,0 +1,202 @@
...
+function previousSong() {
+  const prevIndex = (currentIndex - 1 + songs.length) % songs.length;
+  playSong(songs[prevIndex]);
+  currentIndex = prevIndex;
+}
+
+// playlist display
+function clearSongBgs() {
+    document.querySelectorAll(".playlist__song").forEach(song => {
+        song.classList.remove('selected');
```
bbsmooth, line 128:
```suggestion
        song.classList.remove('selected');
        song.removeAttribute('aria-current');
```

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/scripts.js
```diff
@@ -0,0 +1,202 @@
...
+    const songDivs = document.querySelectorAll(".playlist__song");
+    songDivs[newIndex].classList.add('selected');
+}
+
+
+function playSelectedSong(song, songDiv) {
+    playSong(song);
+    clearSongBgs();
+
+    songDiv.classList.add('selected');
```
bbsmooth, line 142:
```suggestion
    songDiv.classList.add('selected');
    songDiv.setAttribute('aria-current', true);
```

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/scripts.js
```diff
@@ -0,0 +1,202 @@
...
+// playlist display
+function clearSongBgs() {
+    document.querySelectorAll(".playlist__song").forEach(song => {
+        song.classList.remove('selected');
+    });
+}
+
+function updatePlayingSongBkg(newIndex) {
+    const songDivs = document.querySelectorAll(".playlist__song");
+    songDivs[newIndex].classList.add('selected');
```
bbsmooth, line 134:
```suggestion
    songDivs[newIndex].classList.add('selected');
    songDivs[newIndex].setAttribute('aria-current', true);
```

bbsmooth:
I added the `aria-current` attribute to the button of the currently selected song. This will help screen reader users know which song is currently selected as they navigate the play list. I think it's a good accessibility enhancement to add since it is fairly simple.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
```diff
@@ -2,124 +2,181 @@
...
+              <path d="M130 1H0M130 9H0" stroke="#F1BE32" stroke-width="2" />
+            </svg>
+          </div>
+          <div>
+            <button class="playlist__close">X</button>
           </div>
-
-          <ul class="playlist-container"></ul>
         </div>
+        <div id="playlist__songs"></div>
```
bbsmooth, line 177:
```suggestion
        <ul id="playlist__songs" aria-labelled="playlist"></ul>
```

I would suggest we make the playlist an actual list for the benefit of screen reader users. Also give it an accessible name (which will just be the text of the `h2`).

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
```diff
@@ -2,124 +2,181 @@
...
+              width="130"
+              height="10"
+              viewBox="0 0 130 10"
+              fill="none"
+              xmlns="http://www.w3.org/2000/svg"
+            >
+              <path d="M130 1H0M130 9H0" stroke="#F1BE32" stroke-width="2" />
+            </svg>
+          </div>
+          <h2 class="playlist-title">Playlist</h2>
```
bbsmooth, line 161:
```suggestion
          <h2 id="playlist" class="playlist-title">Playlist</h2>
```

Give it an `id` so we can use it to set the accessible name on the playlist itself.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/styles.css
```diff
@@ -1,128 +1,253 @@
...
+*::before {
   box-sizing: border-box;
 }
 
-:root {
-  --main-background-color: #0a0a23;
-  --gray100: #ffffff;
-  --gray200: #e9e8e8;
-  --golden-yellow: #feac32;
+body {
```
bbsmooth, line 15:
```suggestion
ul {
  margin: 0;
}
body {
```

If we use a `ul` for the playlist then we need to get rid of the stupid default margins that browsers add.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/scripts.js
```diff
@@ -0,0 +1,202 @@
...
+  const nextIndex = (currentIndex + 1) % songs.length;
+  playSong(songs[nextIndex]);
+  currentIndex = nextIndex;
+}
+
+function previousSong() {
+  const prevIndex = (currentIndex - 1 + songs.length) % songs.length;
+  playSong(songs[prevIndex]);
+  currentIndex = prevIndex;
+}
```
bbsmooth, line 123:
```suggestion

    if (document.activeElement.classList.contains('playlist__song')) {
        const songDivs = document.querySelectorAll(".playlist__song");
        songDivs[currentIndex].focus();
    }
}
```

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/scripts.js
```diff
@@ -0,0 +1,202 @@
...
+    song.src === audio.src
+}) || songs[currentIndex];
+}
+
+// player controls
+function nextSong() {
+  const nextIndex = (currentIndex + 1) % songs.length;
+  playSong(songs[nextIndex]);
+  currentIndex = nextIndex;
+}
```
bbsmooth, line 117:
```suggestion

    if (document.activeElement.classList.contains('playlist__song')) {
        const songDivs = document.querySelectorAll(".playlist__song");
        songDivs[currentIndex].focus();
    }
}
```

This keeps the keyboard focus indicator in sync with the arrow keys/selected indicator. Right now it is possible to have the focus indicator on a different song than the currently selected song if the user tabs into the playlist. If the keyboard focus is currently on a song in the playlist and the arrow keys are used then the keyboard focus indicator will jump to the currently selected song.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
```diff
@@ -2,124 +2,181 @@
...
-                  fill="#ff0000"
-                  stroke="none"
+                  <path
+                    d="M11.9952 0.75L0.745186 7.24519L11.9952 13.7404V0.75Z"
+                    fill="#F5F6F7"
+                  />
+                </svg>
+              </button>
+              <button class="play">
+                <img src="assets/play.svg" alt="play" />
```
bbsmooth, line 94: To be more accessible, I think we need to add the currently selected song name and artist to the end of the alt text. 

For the pause button I think it's OK if the alt text is just "pause" as the user should already know which song is playing.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/scripts.js
```diff
@@ -0,0 +1,202 @@
...
+
+    songListDiv.innerHTML = songList;
+}
+
+renderSongs(songs);
+
+
+// audio API
+const audio = new Audio();
+
```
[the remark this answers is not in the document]
RafaelDavisH: Thank you @jdwilkin4 🚀

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/scripts.js
RafaelDavisH: Thank you @bbsmooth  - Its looking great and in synced

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-playlist-app/index.html
RafaelDavisH: Yes, I am planning to use this button to display a menu.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/assets/pause.svg
naomi-lgbt, line 1: We will need to host these on our CDN.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/script.js
```diff
@@ -0,0 +1,231 @@
...
+
+// audio API
+const audio = new Audio();
+
+// keepig track of time for each song
+let currentTime = 0;
+
+let currentIndex = 0;
+
+updatePlayingSongBkg(currentIndex);
```
naomi-lgbt, line 62: What does `Bkg` mean? If it's intended as `Background`, we should write out the full word.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/assets/pause.svg
RafaelDavisH: I think it would be best to added to the S3 folder where song and cover art are been hosted. I will have to ask Mrugesh.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/index.html
```diff
@@ -0,0 +1,207 @@
...
+    <link
+      href="https://fonts.googleapis.com/css2?family=Lato&family=Roboto+Mono&display=swap"
+      rel="stylesheet"
+    />
+    <link rel="stylesheet" href="styles.css" />
+    <title>
+      Learn Basic String and Array Methods by Building a Music Player App
+    </title>
+  </head>
+  <body onload="app()">
```
naomi-lgbt, line 19: We have traditionally loaded the script directly, rather than calling an onload function.

I think we should discuss whether we want to teach this approach.

naomi-lgbt (changes requested):
Some quick initial thoughts.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/scripts.js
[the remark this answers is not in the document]
RafaelDavisH: made a new commit to remove the duplicate file

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/script.js
RafaelDavisH: Yes, is short for background. I'll make the change

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/index.html
RafaelDavisH: Instead of using the onload property, we have just call the app function at the end of scripts file. But, I do think it would be good to teach this approach too.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/script.js
```diff
@@ -0,0 +1,231 @@
...
+    if (songListDiv.style.visibility == "hidden") {
+        songListDiv.style.visibility = "visible";
+        songListDiv.style.display = "flex";
+        closeBtn.setAttribute('aria-expanded', true);
+
+    } else {
+        songListDiv.style.visibility = "hidden";
+        songListDiv.style.display = "none";
+        closeBtn.setAttribute('aria-expanded', false);
+    }
```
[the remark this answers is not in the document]
RafaelDavisH: SGTM - You might need to update again. I pushed another commit

RafaelDavisH:
@naomi-lgbt @jdwilkin4 something I forgot to address is what @bbsmooth said about if we should consider to make this project responsive. Its using CSS Grid and Flexbox but more for layout purposes WYAT?

https://github.com/freeCodeCamp/CurriculumExpansion/pull/338#issuecomment-1516826411

naomi-lgbt:
Yes, we should aim for responsive - the project preview window is very small.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/index.html
jdwilkin4: I am not opposed to showing this alternative onload function.
I think we should add a comment above the opening body tag mentioning this is the first time this attribute is used. 
```suggestion
<!-- this is the first time the onload attribute is used. We need to make sure to teach this new concept in the steps -->
  <body onload="app()">
```

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/index.html
```diff
@@ -0,0 +1,207 @@
...
+              aria-expanded="true"
+            >
+              X
+            </button>
+          </div>
+        </div>
+        <ul id="playlist__songs" aria-labelledby="playlist"></ul>
+      </div>
+    </div>
+    <script src="scripts.js"></script>
```
jdwilkin4, line 205: This should be `script.js` so it matches what is in the curriculum editor.
We will also have to change the actual file name too

```suggestion
    <script src="script.js"></script>
```

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/scripts.js
```diff
@@ -0,0 +1,231 @@
...
+    { title: "Walking on Air", cover: "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/quincy-larson-album-art.jpg", artist: "Quincy Larson", duration: "3:25" , src:"https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/walking-on-air.mp3"},
+    { title: "Can't Stop Me. Can't Even Slow Me Down.", cover: "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/quincy-larson-album-art.jpg", artist: "Quincy Larson", duration: "3:52" , src:"https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/cant-stop-me-cant-even-slow-me-down.mp3"},
+    { title: "The Surest Way Out is Through", cover: "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/quincy-larson-album-art.jpg", artist: "Quincy Larson", duration: "3:10" , src:"https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/the-surest-way-out-is-through.mp3"},
+    { title: "Chasing That Feeling", cover: "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/quincy-larson-album-art.jpg", artist: "Quincy Larson", duration: "2:43" , src:"https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/chasing-that-feeling.mp3"},
+  ];
```
jdwilkin4, line 18: This might be a personal preference but I think it would be easier to read with this prettier formatting.

```suggestion
const songs = [
  {
    title: "Scratching The Surface",
    cover:
      "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/quincy-larson-album-art.jpg",
    artist: "Quincy Larson",
    duration: "4:25",
    src: "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/scratching-the-surface.mp3",
  },
  {
    title: "Can't Stay Down",
    cover:
      "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/quincy-larson-album-art.jpg",
    artist: "Quincy Larson",
    duration: "4:15",
    src: "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/cant-stay-down.mp3",
  },
  {
    title: "Still Learning",
    cover:
      "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/quincy-larson-album-art.jpg",
    artist: "Quincy Larson",
    duration: "3:51",
    src: "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/still-learning.mp3",
  },
  {
    title: "Cruising for a Musing",
    cover:
      "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/quincy-larson-album-art.jpg",
    artist: "Quincy Larson",
    duration: "3:34",
    src: "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/cruising-for-a-musing.mp3",
  },
  {
    title: "Never Not Favored",
    cover:
      "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/quincy-larson-album-art.jpg",
    artist: "Quincy Larson",
    duration: "3:35",
    src: "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/never-not-favored.mp3",
  },
  {
    title: "From the Ground Up",
    cover:
      "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/quincy-larson-album-art.jpg",
    artist: "Quincy Larson",
    duration: "3:12",
    src: "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/from-the-ground-up.mp3",
  },
  {
    title: "Walking on Air",
    cover:
      "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/quincy-larson-album-art.jpg",
    artist: "Quincy Larson",
    duration: "3:25",
    src: "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/walking-on-air.mp3",
  },
  {
    title: "Can't Stop Me. Can't Even Slow Me Down.",
    cover:
      "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/quincy-larson-album-art.jpg",
    artist: "Quincy Larson",
    duration: "3:52",
    src: "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/cant-stop-me-cant-even-slow-me-down.mp3",
  },
  {
    title: "The Surest Way Out is Through",
    cover:
      "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/quincy-larson-album-art.jpg",
    artist: "Quincy Larson",
    duration: "3:10",
    src: "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/the-surest-way-out-is-through.mp3",
  },
  {
    title: "Chasing That Feeling",
    cover:
      "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/quincy-larson-album-art.jpg",
    artist: "Quincy Larson",
    duration: "2:43",
    src: "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/chasing-that-feeling.mp3",
  },
];

```

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/scripts.js
```diff
@@ -0,0 +1,231 @@
...
+
+  // song playlist
+const songListDiv = document.getElementById("playlist__songs");
+const closeBtn = document.querySelector(".playlist__close");
```
jdwilkin4, line 22: We haven't really been following the BEM methodology in either of the new JavaScript or new Responsive web design curriculum. 
We have mainly been using kebab case. 

Personally I like to use BEM when working with SASS but if this hasn't been taught in the Responsive web design curriculum then we should stick with kebab case here and change it in the html and css files too

```suggestion
const songListDiv = document.getElementById("playlist-songs");
const closeBtn = document.querySelector(".playlist-close");
```

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/scripts.js
```diff
@@ -0,0 +1,231 @@
...
+
+function playSelectedSong(song, songBtn) {
+    playSong(song);
+    clearSongBgs();
+    songBtn.setAttribute('aria-current', true);
+}
+
+function shuffle() {
+    // Shallow copy done with spread operator
+    let songsCopy = [...songs];
```
jdwilkin4, line 162: I believe this is the first time the spread operator is being introduced. 
So it would be helpful to make a note of that here when we break it down into steps and tests

```suggestion
    // first time the spread operator is being taught. Make sure to introduce with example in the steps
    let songsCopy = [...songs];
```

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/scripts.js
```diff
@@ -0,0 +1,231 @@
...
+  let resetBtn = document.createElement('button');
+  let resetText = document.createTextNode('Reset Playlist');
+  resetBtn.appendChild(resetText);
+  resetBtn.ariaLabel = 'reset playlist';
+  resetBtn.classList.add('player__popup-menu');
+  resetBtn.onclick = function () {
+    app()
+  };
+
+  songs.splice(song,1);
```
jdwilkin4, line 180: I also believe this is the first time where splice is introduced

```suggestion
  // first time splice is introduced. make sure to provide explanation and example in the steps. 
  songs.splice(song,1);
```

jdwilkin4 (changes requested):
I think this project is looking good. 
Left a few comments.

But also, I think we should decide if we want to use the BEM methodology in this project or stick with the kebab case. 
I think it would be good to bring it up in the chat and hear from the other team members 👍

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/script.js
```diff
@@ -0,0 +1,326 @@
...
+      songListDiv.append(resetBtn);
+    }
+  }
+
+  // Event listeners
+  document.querySelector(".play").addEventListener("click", togglePlay);
+  document.querySelector(".pause").addEventListener("click", pauseSong);
+  document.querySelector(".previous").addEventListener("click", previousSong);
+  document.querySelector(".next").addEventListener("click", nextSong);
+  document.querySelector(".shuffle").addEventListener("click", shuffle);
```
jdwilkin4, line 272: I am having hard time getting the shuffle button to work here.

jdwilkin4: While testing locally, if I start a song and then click on the shuffle button then the indicator goes back to the first song but it doesn't look like the list is shuffled

https://github.com/freeCodeCamp/CurriculumExpansion/assets/67210629/a176d111-4809-4f63-95be-6cafd2c14d7d

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/index.html
```diff
@@ -0,0 +1,180 @@
...
+          </div>
+          <div>
+            <button
+              class="playlist-close"
+              aria-label="Playlist"
+              aria-expanded="true"
+            >
+              X
+            </button>
+          </div>
```
larymak, line 173: It doesn't seem like the hamburger at the top has any functionality. Is it possible to move and change it function to alternate with the X button such that when a user clinks the button the playlist closes and the sign changes into the hamburger which in return when clicked can display the playlist?

RafaelDavisH: The top hamburger look-a-line, was just for decoration. I have remove this top 'hamburger' svg and updated the closing playlist button. This new arrow icon is accessible/UX and more intuitive for campers to know what the button/icon is for, a collapsable button.

Thanks for the observation @larymak

scissorsneedfoodtoo (commented):
Took a look earlier and it seems like all of the review comments have been addressed.

Will merge this as-is and will work on a new PR with @JoyShaheb to simplify the project before it's broken into steps.

PR #343, opened 2023-06-29
feat: update calorie counter styling

scissorsneedfoodtoo:
Checklist:

This PR removes the old broken down steps a volunteer helped us with a few years ago. It also, syncs the `index.html`, `styles.css`, and `script.js` files with what we currently have on Learn.

All of the updates to the styling can be found in [this commit](https://github.com/freeCodeCamp/CurriculumExpansion/pull/343/commits/c91fd270dc4df1cacfdc26d3492c81ed3b44d5b8).

Also, here's a link to the styling updates on CodeSandbox: https://codesandbox.io/s/calorie-counter-style-updates-jrhpzk?file=/script.js


javascript-algorithms-and-data-structures/learn-form-validation-by-building-a-calorie-counter/styles.css
```diff
@@ -1,93 +1,94 @@
...
-.btn-outline {
-  color: rgb(40, 167, 69);
-  border: 1px solid rgb(40, 167, 69);
+fieldset,
+label,
+button,
+input,
+select {
+  margin-bottom: 10px;
 }
```
jdwilkin4, line 71: I think we should use a class here instead

jdwilkin4 (changes requested):
I think the style changes look good.
Just made a small comment.

javascript-algorithms-and-data-structures/learn-form-validation-by-building-a-calorie-counter/styles.css
scissorsneedfoodtoo: I played around with the code and added some classes to these elements, but there doesn't seem to be a huge benefit.

We could add class names like `budget-input`, `breakfast`, `lunch`, `dinner`, `exercise`, `entry-name-label`, `entry-name-input`, and so on. However, this could add a few more steps and make the code a bit harder to read. We would also need to modify some of the template strings in the JS to include specific class names.

We could use a utility class instead like `spacing` instead and add that to all the elements, though again it could add more steps and require updating some JS template strings.

Since the page is pretty simple and we want to add a bottom margin to all of these elements uniformly, I think it's easier to stick with type selectors here.

jdwilkin4 (approved):
These changes look good 👍

PR #344, opened 2023-07-26
fix/learn-to-build-music-player-app

JoyShaheb:
Checklist:


jdwilkin4:
Hey @JoyShaheb ! 

We just merged in to accessiblity fixes into the music player app.
Please make sure to incorporate these changes into your PR.

https://github.com/freeCodeCamp/CurriculumExpansion/pull/345
https://github.com/freeCodeCamp/CurriculumExpansion/pull/346

Thanks 👍

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/script.js
```diff
@@ -126,29 +144,23 @@ function renderSongs(array) {
...
-
 // audio API
 const audio = new Audio();
 
-const [playlist,setPlaylist] = useState([...songs]);
-
-const [currentIndex,setCurrentIndex] = useState(0);
+/**
+ * ! These are obsolete
+ */
```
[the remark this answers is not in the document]
JoyShaheb: @RafaelDavisH  please look at the file named newScript, they carry the updated solution, I saved all data in a object and made it significantly simpler

scissorsneedfoodtoo:
This should be ready for review and usability testing now. I believe all of the accessibility concerns that Bruce brought up in his prior issue and reviews should all be addressed here.

Taking another quick look at the code, a few other things we could do to simplify this project are:
- Remove the Google fonts and use browser default fonts
- Use global variables to track state rather than a `userData` object
  - Note: The `userData` object works well, but currently there's a lot of optional chaining. By lifting its properties out into global variables, we could still teach optional chaining in a few cases, and also the spread operator to clone the base array of songs `allSongs` into another array, maybe named `activePlaylist`

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/index.html
```diff
@@ -142,19 +125,10 @@ <h2 class="playlist-title" id="playlist">Playlist</h2>
...
-            </button>
-          </div>
         </div>
-        <ul id="playlist-songs" aria-labelledby="playlist"></ul>
+        <ul id="playlist-songs"></ul>
       </div>
     </div>
-    <script src="script.js"></script>
   </body>
+  <script src="script.js"></script>
```
jdwilkin4, line 133: Any reason why this was moved?

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/script.js
```diff
@@ -1,384 +1,265 @@
- //  Songs array creation
- const songs = [
+const playlistSongs = document.getElementById("playlist-songs");
+const playButton = document.getElementById("play");
+const pauseButton = document.getElementById("pause");
+const nextButton = document.getElementById("next");
+const previousButton = document.getElementById("previous");
+const shuffleButton = document.getElementById("shuffle");
+
+const cover =
```
jdwilkin4, line 8: Why is this cover variable here?
It doesn't look like it is being used anywhere, is it?

My understanding is that we have the image in the HTML and it doesn't change. So it looks like we don't need it in the JS too.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/script.js
```diff
@@ -1,384 +1,265 @@
...
+  "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/quincy-larson-album-art.jpg";
+const allSongs = [
   {
-    id:"0",
+    id: "0",
     title: "Scratching The Surface",
-    cover:
-      "https://s3.amazonaws.com/org.freecodecamp.mp3-player-project/quincy-larson-album-art.jpg",
+    cover,
```
jdwilkin4, line 14: If we are removing the cover variable, then all of these referencing to it can be removed too

jdwilkin4 (changes requested):
tested it locally and it works well on my end. 
Left a couple of questions 👍

jdwilkin4:
@scissorsneedfoodtoo 

> Remove the Google fonts and use browser default fonts

Yeah, I think that is fine. Since this is not about the styling but rather the javascript lessons. 

> Note: The userData object works well, but currently there's a lot of optional chaining.

This might be the only time we teach optional chaining though.
I can't think of another project where it is being used.
So if that is the case, maybe tons of repetition is needed here

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/index.html
scissorsneedfoodtoo: Thanks for catching this. I think this was just an accident on our part.

I'll go ahead and move this back now to just before the closing `</body>` tag.

scissorsneedfoodtoo:
Thank you again for your review, @jdwilkin4, and for your other comments.

> Yeah, I think that is fine. Since this is not about the styling but rather the javascript lessons.

We've had some reports in the past that some learning environments where outside scripts / styling get blocked have trouble with Google Fonts (classes in prisons, restrictive school networks, and so on).

But we have fallbacks in the stylesheet here, so it shouldn't be a problem. If we do get reports that it is an issue, we can always open up another PR on learn to remove those from the seeds on Learn.

> This might be the only time we teach optional chaining though.

True. According to the spreadsheet, none of the other projects teach optional chaining, so this would be a good candidate for introducing those concepts.

We may be able to review optional chaining in other projects like the later `fetch` projects, but we can do that later if necessary.

When this PR is merged and the Music Player project is ready to be broken up into steps, I'll make sure to add optional chaining to the spreadsheet for tracking.

jdwilkin4 (approved):
I did some more testing and I think we are good to go.
let's merge this in so we can start breaking it down into steps and tests 👍

PR #346, opened 2023-07-28
fix(a11y): improve playlist semantics

bbsmooth:
Checklist:

Closes #XXXXX

Technically, you are not supposed to nest `p` and `div` elements inside of a `button`, so I have switched them all to `span`s or removed them completely. I also put `aria-current` on the list item itself (which probably makes more sense). Now we have better semantics and cleaner HTML.


jdwilkin4 (approved):
Looks good 👍

PR #348, opened 2023-08-30
fix: suggested changes for todo app

scissorsneedfoodtoo:
Checklist:

Closes #XXXXX

These updates are based on a recent discussion with Oliver, and should hopefully make it more clear which elements learners are working with while writing JS. I also added a bit of responsiveness as well, and tried to simplify the structure and styling as much as possible.

Also, it adds some code to close the dialog modal with the `.close()` method based on Jessica's suggestion. This should help us get around some platform limitations.

More context here: https://chat.freecodecamp.org/group/front-3-2023?msg=vkafmcPdmbadnDM39

### DIFF
--- a/javascript-algorithms-and-data-structures/learn-debugging-by-building-a-TODO-app/index.html
+++ b/javascript-algorithms-and-data-structures/learn-debugging-by-building-a-TODO-app/index.html
@@ -65,9 +65,9 @@
           </div>
           <div class="task-form-footer">
             <button
-              id="task-btn"
+              id="add-or-update-task-btn"
               class="btn large-btn"
-              type="submit"
+              type="button"
               aria-label="add task"
             >
               Add Task
--- a/javascript-algorithms-and-data-structures/learn-debugging-by-building-a-TODO-app/index.html
+++ b/javascript-algorithms-and-data-structures/learn-debugging-by-building-a-TODO-app/index.html
@@ -14,7 +14,7 @@
         <button id="open-task-form-btn" class="btn large-btn">
           Add New Task
         </button>
-        <form class="task-form" id="task-form">
+        <form class="task-form hidden" id="task-form">
           <div class="task-form-header">
             <button
               id="close-task-form-btn"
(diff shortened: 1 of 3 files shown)


javascript-algorithms-and-data-structures/learn-debugging-by-building-a-TODO-app/index.html
```diff
@@ -4,103 +4,92 @@
...
             </button>
-            <button class="btn" id="confirm-btn" value="default">Exit</button>
           </div>
         </form>
-      </dialog>
-
-      <div id="tasks-container"></div>
+        <dialog id="confirm-close-dialog">
+          <form method="dialog">
+            <p class="discard-message-text">Discard unsaved changes?</p>
```
jdwilkin4, line 79: I think we should update the language because technically a user could close the modal without changing anything first.


Screenshot 2023-08-30 at 9 02 14 AM

How about something like this?

Screenshot 2023-08-30 at 9 02 42 AM

Ksound22:
Hey Kris 👋🏽

Thanks a lot for looking into this.

https://github.com/freeCodeCamp/CurriculumExpansion/assets/65571316/dfdfe2bd-e70e-4420-8116-c6658286dda5

I did some testing on my end. The modal now closes, but the initial error in the console persists.

If that won't cause any further issues then we are good to go.

scissorsneedfoodtoo:
Hey @jdwilkin4 and @Ksound22, thanks for your patience. This should be ready for testing now.

@jdwilkin4, there are now some checks for whether the inputs have values in them when adding a new task, or whether the values have been updated when editing a task, so we may be able to leave the modal text as is. But if you'd like to update the text from "Discard current changes?" to something else, please let me know.

@Ksound22, we should be able to ignore that error for now. That's coming from the check in `client/src/templates/Challenges/utils/frame.ts`, and we should be able to add optional chaining there to prevent the error for this project without causing issues for existing projects.

Ksound22:
Those are some good improvements, Kris.

Things are good to go on my end!

jdwilkin4 (approved):
tested it out locally and everything is working on my end.
I vote we merge this in so we can resume breaking it down into steps and tests

PR #349, opened 2023-09-13
fix: music player improvements

scissorsneedfoodtoo:
Checklist:

This PR adds a couple of important improvements like playing the next song after the current one is finished. Also, reseting the player state and HTML attributes for accessibility in certain states (the final song finishes playing, songs are shuffled, and so on).

Other changes include reordering the order of the functions, renaming functions for clarity (`playNextSong` rather than `nextSong`), and using numbers for song ids instead of strings to simplify the code a bit.

It also adds some comments that may help while breaking the project down into steps.

### DIFF
--- a/javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/script.js
+++ b/javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/script.js
@@ -96,6 +96,7 @@ const deleteSong = (id) => {
 
   userData.songs = userData?.songs.filter((song) => song.id !== id.toString());
   renderSongs(userData?.songs);
+  songHighlighter();
   setPlayButtonAccessibleText();
 
   if (userData.songs.length === 0) {
@@ -111,31 +112,36 @@ const deleteSong = (id) => {
       userData.songs = allSongs;
 
       renderSongs(userData?.songs);
+      setPlayButtonAccessibleText();
       resetButton.remove();
     });
   }
 };
 
-const setPlayerDisplay = (title, artist) => {
+const setPlayerDisplay = () => {
   const playingSong = document.getElementById("player-song-title");
   const songArtist = document.getElementById("player-song-artist");
+  const currentTitle = userData?.currentSong?.title;
+  const currentArtist = userData?.currentSong?.artist;
 
-  playingSong.textContent = title ? title : "";
-  songArtist.textContent = artist ? artist : "";
+  playingSong.textContent = currentTitle ? currentTitle : "";
+  songArtist.textContent = currentArtist ? currentArtist : "";
 };
 
-const songHighlighter = (id) => {
+const songHighlighter = () => {
   const playlistSongElements = document.querySelectorAll(".playlist-song");
-  const songToHighlight = document.getElementById(`song-${id}`);
+  const songToHighlight = document.getElementById(`song-${userData?.currentSong?.id}`);
 
-  playlistSongElements.forEach((song) => {
-    song.removeAttribute("aria-current");
+  // Introduce the .forEach() method
+  playlistSongElements.forEach((songEl) => {
+    songEl.removeAttribute("aria-current");
   });
 
-  songToHighlight.setAttribute("aria-current", "true");
+  if (songToHighlight) songToHighlight.setAttribute("aria-current", "true");
 };
 
 const renderSongs = (array) => {
+  // Introduce the .map() method
   const songsHTML = array
     .map((song) => {
       return `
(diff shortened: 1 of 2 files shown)


bbsmooth:
> playing the next song after the current one is finished

This seems like a nice feature to add since that seems like the default behavior for most music players. One concern is that we currently have no way to advance to the end of a song, so in order to test this you have to wait for the entire song to finish. OK, not the end of the world, but somewhat annoying, especially if you want to test it for more than one song. Perhaps we could use just stubs for the songs (like just the first 10-15 seconds of the song) instead of the entire song?

Also, why not make it wrap so that after the last song in the list has played it starts over at the beginning?

scissorsneedfoodtoo:
Hi @bbsmooth, thanks for your patience and for your helpful feedback.

I just added some short tracks to make it easier to test the feature where the next track is played when the current one is finished.

About the removal of the playback wrapping, we talked about this in a meeting sometime back. Apparently a lot of music players only do that when repeat mode is on. I've only tested this on Spotify and YouTube Music, but that seems to be the case. Also, we were looking for ways to simplify the project slightly. Rafa's code to either play the next / previous track, or wrap around to the first / last song when the next or previous button was pressed worked really well, but seemed pretty complex for the third project in the new JS curriculum.

@JoyShaheb, could you review this and merge this if you're happy with the changes?

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/script.js
```diff
@@ -5,97 +5,182 @@ const nextButton = document.getElementById("next");
 const previousButton = document.getElementById("previous");
 const shuffleButton = document.getElementById("shuffle");
 
+// Short mp3s for testing
```
jdwilkin4, line 8: since this look like it was only used for testing locally, I think we should revert the allSongs list back to its original state

scissorsneedfoodtoo: Thanks for bringing this up @jdwilkin4.

I pushed a commit to have these test mp3s commented out by default.

bbsmooth: What if campers want to test this themselves as they are building the project? I think we should leave in a few very short mp3s so that they can easily do this.

RafaelDavisH: A minor thing: The dashed outline stays put and does not follow to the next song. It's unsynced with song playing.

RafaelDavisH:
A minor thing: The dashed outline stays put and does not follow to the next song. It's unsynced with song playing.

javascript-algorithms-and-data-structures/learn-basic-string-and-array-methods-by-building-a-music-player-app/script.js
scissorsneedfoodtoo: Thanks for this helpful suggestion, @bbsmooth.

I was working on this project with @RafaelDavisH today and he offered to create some short 10 second MP3 files we can use for some steps in this project.

I'm thinking that, for the steps where learners will write code to play the next song once the current song is finished, we can update the `allSongs` array in the seed for them with the shorter files for testing, and mention it in the description. Once learners are finished coding and testing that feature, we can revert the seed back to the original `allSongs` array with the longer tracks.

scissorsneedfoodtoo:
@Ksound22 and @larymak reviewed this yesterday and say everything looks good.

Unfortunately they don't have the ability to merge this PR, so I'll go ahead and do it on their behalf so they can continue breaking this project into steps.

bbsmooth:
> The dashed outline stays put and does not follow to the next song. It's unsynced with song playing.

Sorry for the late reply, but I have some reservations about whether this is something we want to do. In general it is not a good idea to change the keyboard focus unless a user initiated action causes the change. In this case, the player is making a change, not the user. The keyboard focus/dashed line does not have to be in sync at all times with the song currently playing in the list. When a song is playing the user is free to move the keyboard focus somewhere other than the song. Forcing it back to the song list when the next song starts seems unnecessary. For example, perhaps they wanted to leave the focus on the next button so they could easily jump to the next song if they didn't like the current one? Granted, with only 10 songs in the play list it wouldn't be much of a hassle to tab back to the next button. But what if there were 50, or 100 songs in the playlist?

Also, this change is causing the keyboard focus to jump to the song in the play list when you use the prev or next buttons at the top, which is definitely not something we want to do. So I guess I am voting that we revert this change and allow the keyboard focus to stay where it is when a song changes.

scissorsneedfoodtoo:
Thanks for your detailed feedback @bbsmooth. Those are very good points, especially if there's a long playlist.

I also wasn't sure about that recent change with the focus, but apparently that was the behavior before.

I can create a quick PR to revert that. Glad to know that we'll be going back to the expected focus behavior, and be making the project a bit simpler as well.

PR #351, opened 2023-10-16
feat(curriculum): add binary search tree

zairahira:
Checklist:

Closes #XXXXX

Adding a new project to the Python curriculum.

### DIFF
--- a/scientific-computing-with-python/learn-tree-traversal-by-building-a-binary-search-tree/binary-search-tree.py
+++ b/scientific-computing-with-python/learn-tree-traversal-by-building-a-binary-search-tree/binary-search-tree.py
@@ -69,13 +69,9 @@ class BinarySearchTree:
 
 # Example usage:
 bst = BinarySearchTree()
-bst.insert(50)
-bst.insert(30)
-bst.insert(20)
-bst.insert(40)
-bst.insert(70)
-bst.insert(60)
-bst.insert(80)
+nodes = [50, 30, 20, 40, 70, 60, 80]
+for node in nodes:
+    bst.insert(node)
 
 print("Inorder traversal:", bst.inorder_traversal())  # Should print [20, 30, 40, 50, 60, 70, 80]
 
--- a/scientific-computing-with-python/learn-tree-traversal-by-building-a-binary-search-tree/binary-search-tree.py
+++ b/scientific-computing-with-python/learn-tree-traversal-by-building-a-binary-search-tree/binary-search-tree.py
@@ -11,69 +11,72 @@ class BinarySearchTree:
     def insert(self, key):
         self.root = self._insert(self.root, key)
 
-    def _insert(self, root, key):
-        if root is None:
+    def _insert(self, node, key):
+        if node is None:
             return TreeNode(key)
-        if key < root.key:
-            root.left = self._insert(root.left, key)
-        else:
-            root.right = self._insert(root.right, key)
-        return root
+        if key < node.key:
+            node.left = self._insert(node.left, key)
+        elif key > node.key:
+            node.right = self._insert(node.right, key)
+        return node
 
     def search(self, key):
         return self._search(self.root, key)
 
-    def _search(self, root, key):
-        if root is None or root.key == key:
-            return root
-        if key < root.key:
-            return self._search(root.left, key)
-        return self._search(root.right, key)
+    def _search(self, node, key):
+        if node is None or node.key == key:
+            return node
+        if key < node.key:
+            return self._search(node.left, key)
+        return self._search(node.right, key)
 
     def delete(self, key):
         self.root = self._delete(self.root, key)
 
-    def _delete(self, root, key):
-        if root is None:
-            return root
-        if key < root.key:
-            root.left = self._delete(root.left, key)
-        elif key > root.key:
-            root.right = self._delete(root.right, key)
+    def _delete(self, node, key):
+        if node is None:
+            return node
+        if key < node.key:
+            node.left = self._delete(node.left, key)
+        elif key > node.key:
+            node.right = self._delete(node.right, key)
         else:
-            # Node with one child or no child
-            if root.left is None:
-                return root.right
-            elif root.right is None:
-                return root.left
-            # Node with two children, get the in-order successor (smallest in the right subtree)
-            root.key = self._min_value_node(root.right).key
-            root.right = self._delete(root.right, root.key)
-        return root
+            if node.left is None:
+                return node.right
+            elif node.right is None:
+                return node.left
+            node.key = self._min_value(node.right)
+            node.right = self._delete(node.right, node.key)
+        return node
 
-    def _min_value_node(self, node):
+    def _min_value(self, node):
         while node.left is not None:
             node = node.left
-        return node
+        return node.key
 
     def inorder_traversal(self):
         result = []
         self._inorder_traversal(self.root, result)
         return result
 
-    def _inorder_traversal(self, root, result):
-        if root:
-            self._inorder_traversal(root.left, result)
-            result.append(root.key)
-            self._inorder_traversal(root.right, result)
+    def _inorder_traversal(self, node, result):
+        if node:
+            self._inorder_traversal(node.left, result)
+            result.append(node.key)
+            self._inorder_traversal(node.right, result)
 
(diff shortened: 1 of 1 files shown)


scientific-computing-with-python/learn-tree-traversal-by-building-a-binary-search-tree/binary-search-tree.py
```diff
@@ -0,0 +1,83 @@
...
+
+# Example usage:
+bst = BinarySearchTree()
+bst.insert(50)
+bst.insert(30)
+bst.insert(20)
+bst.insert(40)
+bst.insert(70)
+bst.insert(60)
+bst.insert(80)
```
Dario-DC, line 78: We could make it more concise with a loop
```suggestion
nodes = [50, 30, 20, 40, 70, 60, 80]
for node in nodes:
    bst.insert(node)
```

scientific-computing-with-python/learn-tree-traversal-by-building-a-binary-search-tree/binary-search-tree.py
```diff
@@ -0,0 +1,83 @@
+class TreeNode:
+    def __init__(self, key):
+        self.key = key
```
gikf, line 3: Seems like `number` or `value` might be better name, instead of `key`. Example at the bottom uses only numbers, but technically there shouldn't be issues with using it for any class that supports comparison between objects from that class.

scientific-computing-with-python/learn-tree-traversal-by-building-a-binary-search-tree/binary-search-tree.py
```diff
@@ -0,0 +1,83 @@
...
+        self.root = self._insert(self.root, key)
+
+    def _insert(self, root, key):
+        if root is None:
+            return TreeNode(key)
+        if key < root.key:
+            root.left = self._insert(root.left, key)
+        else:
+            root.right = self._insert(root.right, key)
+        return root
```
gikf, line 21: `_insert` is a static method - it doesn't use any instance or class attributes. This applies also to other `_x` methods defined in `BinarySearchTree`

```suggestion
    def insert(self, key):
        self.root = BinarySearchTree._insert(self.root, key)

    @staticmethod
    def _insert(root, key):
        if root is None:
            return TreeNode(key)
        if key < root.key:
            root.left = BinarySearchTree._insert(root.left, key)
        else:
            root.right = BinarySearchTree._insert(root.right, key)
        return root
```

scientific-computing-with-python/learn-tree-traversal-by-building-a-binary-search-tree/binary-search-tree.py
```diff
@@ -0,0 +1,83 @@
...
+        self.key = key
+        self.left = None
+        self.right = None
+
+class BinarySearchTree:
+    def __init__(self):
+        self.root = None
+
+    def insert(self, key):
+        self.root = self._insert(self.root, key)
```
gikf, line 12: If I'm not mistaken, structuring inserts and deletions like this will actually re-assign multiple nodes, even if that's not needed. The bigger tree, the more re-assigns will occur. That's a bit pointless.

scientific-computing-with-python/learn-tree-traversal-by-building-a-binary-search-tree/binary-search-tree.py
```diff
@@ -0,0 +1,83 @@
...
+    def inorder_traversal(self):
+        result = []
+        self._inorder_traversal(self.root, result)
+        return result
+
+    def _inorder_traversal(self, root, result):
+        if root:
+            self._inorder_traversal(root.left, result)
+            result.append(root.key)
+            self._inorder_traversal(root.right, result)
```
gikf, line 68: Slight shuffling things around, to omit the need to pass list around. Instead `_inorder_traversal` static method returns ordered list.

```suggestion
    def inorder_traversal(self):
        return BinarySearchTree._inorder_traversal(self.root)

    @staticmethod
    def _inorder_traversal(root):
        result = []
        if root:
            result.extend(BinarySearchTree._inorder_traversal(root.left))
            result.append(root.key)
            result.extend(BinarySearchTree._inorder_traversal(root.right))
        return result
```

scientific-computing-with-python/learn-tree-traversal-by-building-a-binary-search-tree/binary-search-tree.py
```diff
@@ -0,0 +1,83 @@
...
+        else:
+            root.right = self._insert(root.right, key)
+        return root
+
+    def search(self, key):
+        return self._search(self.root, key)
+
+    def _search(self, root, key):
+        if root is None or root.key == key:
+            return root
```
gikf, line 28: Shouldn't search return `True` or `False`, instead of `None` or whatever `key` will be?

scientific-computing-with-python/learn-tree-traversal-by-building-a-binary-search-tree/binary-search-tree.py
Dario-DC: I think using `key` is fine. In the challenge step description we can mention that, although only numerical keys will be used in the project.

Dario-DC (commented):
I have one major doubt: **if** this is going to be the first project explaining classes, maybe having a bunch of `x` methods and their correspondents `_x` static methods could be confusing.

scientific-computing-with-python/learn-tree-traversal-by-building-a-binary-search-tree/binary-search-tree.py
```diff
@@ -0,0 +1,82 @@
...
+        elif key > node.key:
+            node.right = self._insert(node.right, key)
+        return node
+
+    def search(self, key):
+        return self._search(self.root, key)
+
+    def _search(self, node, key):
+        if node is None or node.key == key:
+            return node
```
Dario-DC, line 28: If you want to return the node instead of True/False, I suggest to add a `__str__` method to the `TreeNode` class:
```py
def __str__(self):
return str(self.key)
```
Otherwise you will see something like `Search for 40: <__main__.TreeNode object at 0x7fee39b43b50>` when you print the result.

scientific-computing-with-python/learn-tree-traversal-by-building-a-binary-search-tree/binary-search-tree.py
```diff
@@ -0,0 +1,82 @@
...
+for node in nodes:
+    bst.insert(node)
+
+print("Inorder traversal:", bst.inorder_traversal())
+
+print("Search for 40:", bst.search(40))
+print("Search for 30:", bst.search(30))
+
+bst.delete(40)
+print("Inorder traversal after deleting 40:", bst.inorder_traversal())
```
Dario-DC, line 82: Maybe you can search again for 40 after the deletion, to show the output when the node is not found.

scientific-computing-with-python/learn-tree-traversal-by-building-a-binary-search-tree/binary-search-tree.py
gikf: For `insert` I was thinking about something like this:
```python
    def insert(self, key):
        new_node = TreeNode(key)
        if self.root is None:
            self.root = new_node
            return
        self._insert(self.root, new_node)

    def _insert(self, parent_node, new_node):
        if new_node.key < parent_node.key:
            if parent_node.left is None:
                parent_node.left = new_node
                return
            self._insert(parent_node.left, new_node)
        elif new_node.key > parent_node.key:
            if parent_node.right is None:
                parent_node.right = new_node
                return
            self._insert(parent_node.right, new_node)
```

Version without recursion:
```python
    def insert(self, key):
        new_node = TreeNode(key)
        if self.root is None:
            self.root = new_node
            return

        parent_node = self.root
        while True:
            if new_node.key < parent_node.key:
                if parent_node.left is None:
                    parent_node.left = new_node
                    return
                parent_node = parent_node.left
            elif new_node.key > parent_node.key:
                if parent_node.right is None:
                    parent_node.right = new_node
                    return
                parent_node = parent_node.right
            else:
                break
```

PR #359, opened 2023-11-06
Fix/improve-platformer

ieahleen:
I have started making adjustments. For now:
 - the player will only catch the checkpoint if it's on the checkpoint, not at any point to its right
 - the player will not disappear on the left side of the screen, it stays at a max of its width distance from the side

If there are any other points that you think need to be touched upon please tell me


javascript-algorithms-and-data-structures/Learn-Intermediate-OOP-by-Building-a-Platformer-Game/script.js
```diff
@@ -42,9 +41,13 @@ class Player {
     } else {
       this.velocity.y = 0;
     }
+
     if (this.position.x < this.width) {
       this.position.x = this.width;
     }
+    if (this.position.x >= canvas.width - 2*this.width) {
+      this.position.x = canvas.width - 2*this.width
+    }
```
jdwilkin4, line 50:
```suggestion
    if (this.position.x >= canvas.width - 2 * this.width) {
      this.position.x = canvas.width - 2 * this.width
    }
```

jdwilkin4 (commented):
Small nitpick but otherwise the changes look good 👍

zairahira:
I have tested the following changes, and they look good to me:

The player is catching the checkpoint exactly at the checkpoint, and not missing it.
The player is not disappearing on the right side of the screen now.

Dario-DC:
One thing might need some improvements is this:

https://github.com/freeCodeCamp/CurriculumExpansion/assets/105294544/8ba51d8b-e135-41cb-9c96-5900d6f01a5a

The player shouldn't disappear below the lower platform and outside the screen.

bbsmooth:
As far as game play goes, I can skip the first two checkpoints and just hit the third one and "win" the game. Not sure if this is something you are concerned about. Also, might be nice to add some things that you don't want to make contact with, but perhaps that would make this project too long.

Concerning accessibility, I am not a gamer and thus I do not have any experience with accessibility and gaming, so I can only offer simple feedback here. I doubt a blind person is going to be able to play this game. I would assume someone using voice control would have a hard time playing it. And people who can't use a physical keyboard would also struggle to play it. Since I'm not familiar with how games typically deal with these types of accessibility issues I don't have any specific recommendations. And it is definitely a possibility that the game may never be able to be made accessible to people with certain disabilities. My only actionable recommendation at this point would be to add a disclaimer on the first step acknowledging that the game has accessibility issues and not everyone will be able to play it.

gikf:
> The player shouldn't disappear below the lower platform and outside the screen.

Other than collisions complications there's one more area that should be considered when fixing this - height of the game field. Player, platforms and checkpoints are drawn at absolute coordinates, regardless of the height of the field. With small enough window, initially might not be displayed anything at all. Maybe use fixed height, when window is not bigger than some value?

ieahleen:
Considering your feedback I guess these are the things that I should focus on:
- [ ] the player shouldn't disappear below the edge of the screen
- [ ] the player shouldn't collide *like that* with the platform and be yeeted below
- [ ] the player should have to get to the checkpoints in order
- [ ] height of the canvas is fixed --- smaller screens should show the game anyway
- [ ] the game is not accessible at all, nor playable on touch screen

jdwilkin4:
@ieahleen 

> the game is not accessible at all, and only playable via keyboard

For this last part on accessibility, it will take a lot more time in research and planning to make changes to make this more accessible. I did some research online and there seems to be a lot of discussion around this but very little practical solutions. The research phase will probably take a while but would be good to look into.

This will probably require a separate PR for accessibility changes IMO.

My advice would be to add the disclaimer in step 1 about accessibility for the beta launch.

Then after the beta launch, we can work on the accessibility portion

ieahleen:
@jdwilkin4 thank you, I will try to have this done before next meeting, so that it can then be reviewed and we can then rework the steps to reflect the changes here

bbsmooth:
> * the game is not accessible at all, and only playable via keyboard

Let me just clarify here a little. The only method of operation that WCAG requires is the keyboard, which this game meets, so that in itself is not technically an accessibility issue. My comment above was that the nature of how the keyboard is used in this game could cause problems for some people because of how arrow keys work together. And ya, we have plenty of people going through fCC with touch devices, so a keyboard-only game might be a bit of a bummer for them to work through. But that's just a usability issue, not an accessibility issue.

I imagine making this fairly accessible is going to take a lot of effort and add significantly more complexity, something I'm guessing you wouldn't want to add to this project. I suppose it is possible to learn from the code itself without being able to actually play the game, but you'll want to make sure that the steps explain enough so that everything is obvious and can be learned without having to see or play the game.

I'm thinking the disclaimer should probably include at least these two thing:
- An acknowledgment that the game itself will be unusable and inaccessible to some people
- A belief that you can still learn from the code itself without needing to play the game

ieahleen:
ok, I'm back.

The conflicts between platforms and checkpoints and canvas size is because the canvas is drawn with
```javascript
canvas.width = innerWidth;
canvas.height = innerHeight;
```
and checkpoints and platforms are drawn with a fixed height.
Possibilities to adjust this:
- set platforms and checkpoints to a height calculated based on canvas.height
- or set canvas to a fixed height
it's kind of difficult to draw it with a perfect balance, because at this point it's maybe better to draw everything proportional, maybe below 500 height, everything proportional, above that fixed height of objects?

ieahleen:
Ok, I made a commit that could work with resizing the canvas and objects. Still not perfect, btw. Now the thing missing is fixing the collision that yeet the player below the canvas.

javascript-algorithms-and-data-structures/Learn-Intermediate-OOP-by-Building-a-Platformer-Game/script.js
```diff
@@ -1,39 +1,42 @@
...
+const ctx = canvas.getContext('2d');
 canvas.width = innerWidth;
 canvas.height = innerHeight;
 const gravity = 0.5;
-let activeCheckpointCollisionDetection = true;
+let isCheckpointCollisionDetectionActive = true;
+
+function proportionalSize(s) {
+  return innerHeight < 500 ? Math.ceil((s / 500) * innerHeight) : s;
+}
```
jdwilkin4, line 14: I think something like size would be a better parameter name 

```suggestion
function proportionalSize(size) {
  return innerHeight < 500 ? Math.ceil((size / 500) * innerHeight) : size;
}
```

jdwilkin4 (changes requested):
Things look good. 
Left one comment.

This first PR can just focus on these couple of changes and then it is totally fine to open separate PR's to address the other changes. 👍

Dario-DC (approved):

jdwilkin4:
I am going to go ahead and merge this in since it has two approvals.
Then for other changes we just create separate PR's for those 👍

PR #365, opened 2023-11-24
Add expense tracker project

tomitokko:
Checklist:

Closes #XXXXX

Added the expense tracker project

### DIFF
--- a/scientific-computing-with-python/learn-lambda-functions-by-building-a-expense-tracker/expense-tracker.py
+++ b/scientific-computing-with-python/learn-lambda-functions-by-building-a-expense-tracker/expense-tracker.py
@@ -1,0 +1,54 @@
+expenses = []
+
+def add_expense(amount, category):
+    expenses.append({"amount": amount, "category": category})
+
+def list_expenses():
+    for expense in expenses:
+        print(f"Amount: {expense['amount']}, Category: {expense['category']}")
+
+def total_expenses():
+    return sum(map(lambda expense: expense['amount'], expenses))
+
+def filter_expenses_by_category(category):
+    filtered_expenses = filter(lambda expense: expense['category'] == category, expenses)
+    for expense in filtered_expenses:
+        print(f"Amount: {expense['amount']}, Category: {expense['category']}")
+
+def main():
+    while True:
+        print("\nExpense Tracker")
+        print("1. Add an expense")
+        print("2. List all expenses")
+        print("3. Show total expenses")
+        print("4. Filter expenses by category")
+        print("5. Exit")
+
+        choice = input("Enter your choice: ")
+
+        if choice == '1':
+            amount = float(input("Enter amount: "))
+            category = input("Enter category: ")
+            add_expense(amount, category)
+
+        elif choice == '2':
+            print("\nAll Expenses:")
+            list_expenses()
+
+        elif choice == '3':
+            print("\nTotal Expenses: ", total_expenses())
+
+        elif choice == '4':
+            category = input("Enter category to filter: ")
+            print(f"\nExpenses for {category}:")
+            filter_expenses_by_category(category)
+
+        elif choice == '5':
+            print("Exiting the program.")
+            break
+
+        else:
+            print("Invalid choice. Please try again.")
+
+if __name__ == "__main__":
+    main()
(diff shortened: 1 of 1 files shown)


scientific-computing-with-python/learn-lambda-functions-by-building-a-expense-tracker/expense-tracker.py
```diff
@@ -0,0 +1,54 @@
+expenses = []
+
+def add_expense(amount, category):
+    expenses.append({"amount": amount, "category": category})
+
+def list_expenses():
```
gikf, line 6: I'd change `list_expenses` to `print_expenses` and make it accept one parameter with expenses to print. That way it can be used for all printing list of expenses needs in the project, instead of having printing in multiple places.

```suggestion
def print_expenses(expenses):
```

scientific-computing-with-python/learn-lambda-functions-by-building-a-expense-tracker/expense-tracker.py
```diff
@@ -0,0 +1,54 @@
...
+        choice = input("Enter your choice: ")
+
+        if choice == '1':
+            amount = float(input("Enter amount: "))
+            category = input("Enter category: ")
+            add_expense(amount, category)
+
+        elif choice == '2':
+            print("\nAll Expenses:")
+            list_expenses()
```
gikf, line 36:
```suggestion
            print_expenses(expenses)
```

scientific-computing-with-python/learn-lambda-functions-by-building-a-expense-tracker/expense-tracker.py
```diff
@@ -0,0 +1,54 @@
...
+    for expense in expenses:
+        print(f"Amount: {expense['amount']}, Category: {expense['category']}")
+
+def total_expenses():
+    return sum(map(lambda expense: expense['amount'], expenses))
+
+def filter_expenses_by_category(category):
+    filtered_expenses = filter(lambda expense: expense['category'] == category, expenses)
+    for expense in filtered_expenses:
+        print(f"Amount: {expense['amount']}, Category: {expense['category']}")
```
gikf, line 16:
```suggestion
    return filter(lambda expense: expense['category'] == category, expenses)
```
After that `filter_expenses_by_category` only filters expenses.

scientific-computing-with-python/learn-lambda-functions-by-building-a-expense-tracker/expense-tracker.py
```diff
@@ -0,0 +1,54 @@
...
+            print("\nAll Expenses:")
+            list_expenses()
+
+        elif choice == '3':
+            print("\nTotal Expenses: ", total_expenses())
+
+        elif choice == '4':
+            category = input("Enter category to filter: ")
+            print(f"\nExpenses for {category}:")
+            filter_expenses_by_category(category)
```
gikf, line 44:
```suggestion
            expenses_from_category = filter_expenses_by_category(category)
            print_expenses(expenses_from_category)
```
And filtered expenses are printed by `print_expenses`.

It might be also a good idea to print total of the filtered expenses. What would mean modifying `total_expenses` to use passed in list.

scientific-computing-with-python/learn-lambda-functions-by-building-a-expense-tracker/expense-tracker.py
tomitokko: I just tested all of this, and this is a great suggestion, thanks.

gikf:
There's one thing left that can make usage of the script/module a bit awkward. There's `if __name__ == "__main__":` which shows that code can be used as a standalone script, or be imported from other module.  No problems when running it as a single script. However single `expenses` list makes usage as an import problematic, as for example, it's not going to be possible to create two different lists for two persons.

Therefore I'd recommend to either make it work only as a script (by removing mentioned `if` and just calling `main()` at the end), or further modify code to give ability to fully work as script and as an import (mostly that means not using global `expenses`, but working with list that would be passed to each function).

tomitokko:
> There's one thing left that can make usage of the script/module a bit awkward. There's `if __name__ == "__main__":` which shows that code can be used as a standalone script, or be imported from other module. No problems when running it as a single script. However single `expenses` list makes usage as an import problematic, as for example, it's not going to be possible to create two different lists for two persons.
> 
> Therefore I'd recommend to either make it work only as a script (by removing mentioned `if` and just calling `main()` at the end), or further modify code to give ability to fully work as script and as an import (mostly that means not using global `expenses`, but working with list that would be passed to each function).

This is a great suggestion, thanks. Instead of just removing the `if` I have refactored the whole code and it should be good now.

PR #371, opened 2024-02-13
feat: add vector space prototype

Dario-DC:
Checklist:

A possible project to add to the Python curriculum.

### DIFF
--- a/scientific-computing-with-python/learn-special-methods-by-building-a-vector-space/vectors.py
+++ b/scientific-computing-with-python/learn-special-methods-by-building-a-vector-space/vectors.py
@@ -1,0 +1,87 @@
+class Vector:
+    space = 'R3'
+
+    def __init__(self, x, y, z):
+        self.x = x
+        self.y = y
+        self.z = z
+        self.norm = (x**2 + y**2 + z**2)**0.5
+
+    def __str__(self):
+        return str((self.x, self.y, self.z))
+    
+    def __repr__(self):
+        return f'Vector({self.x}, {self.y}, {self.z})'
+
+    def __add__(self, other):        
+        x = self.x + other.x
+        y = self.y + other.y
+        z = self.z + other.z
+        v = Vector(x, y, z)
+        return v
+    
+    def __sub__(self, other):
+        x = self.x - other.x
+        y = self.y - other.y
+        z = self.z - other.z
+        v = Vector(x, y, z)
+        return v
+    
+    def __mul__(self, other):
+        # Scalar multiplication
+        if isinstance(other, int) or isinstance(other, float):
+            self.x *= other
+            self.y *= other
+            self.z *= other
+            self.norm *= other
+            return self
+        # Scalar product
+        elif isinstance(other, Vector):
+            x = self.x * other.x
+            y = self.y * other.y
+            z = self.z * other.z
+            scalar = (x**2 + y**2 + z**2)**0.5
+            return scalar
+        
+    def cross(self, other):
+        # Cross product
+        x = self.y * other.z - self.z * other.y
+        y = self.z * other.x - self.x * other.z
+        z = self.x * other.y - self.y * other.x
+        v = Vector(x, y, z)
+        return v
+    
+    def __eq__(self, other):
+        if self.x == other.x \
+        and self.y == other.y \
+        and self.z == other.z:
+            return True
+        return False
+    
+    def __ne__(self, other):
+        return not self == other
+    
+    def __lt__(self, other):
+        return self.norm < other.norm
+    
+    def __gt__(self, other):
+        return not self < other
+    
+    def __le__(self, other):
+        return self.norm <= other.norm
+    
+    def __ge__(self, other):
+        return not self <= other
+
+
+# Example usage
+v1 = Vector(3, 2, 2)
+v2 = Vector(4.7, 1, 1.5)
+print('v1 =', v1)
+print('v2 =', v2)
+v3 = v1 + v2
+print('v1 + v2 =', v3)
+n = v1 * v2
+print('v1 * v2 =', n)
+v4 = v1.cross(v2)
+print('v1 x v2 =', v4)
--- a/scientific-computing-with-python/learn-special-methods-by-building-a-vector-space/vectors.py
+++ b/scientific-computing-with-python/learn-special-methods-by-building-a-vector-space/vectors.py
@@ -16,13 +16,13 @@ class R2Vector:
     
     def __add__(self, other):
         if not isinstance(other, self.__class__):
-            return NotImplemented
+            raise TypeError(f"Addition is not defined between '{self.__class__.__name__}' and '{other.__class__.__name__}'")
         kwargs = {i: self.__getattribute__(i) + other.__getattribute__(i) for i in self.__dict__}
         return self.__class__(**kwargs)
     
     def __sub__(self, other):
         if not isinstance(other, self.__class__):
-            return NotImplemented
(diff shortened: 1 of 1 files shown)


scientific-computing-with-python/learn-special-methods-by-building-a-vector-space/vectors.py
```diff
@@ -0,0 +1,87 @@
...
+    def __sub__(self, other):
+        x = self.x - other.x
+        y = self.y - other.y
+        z = self.z - other.z
+        v = Vector(x, y, z)
+        return v
+    
+    def __mul__(self, other):
+        # Scalar multiplication
+        if isinstance(other, int) or isinstance(other, float):
```
gikf, line 32:
```suggestion
        if isinstance(other, (float, int)):
```
A bit simpler.

scientific-computing-with-python/learn-special-methods-by-building-a-vector-space/vectors.py
```diff
@@ -0,0 +1,87 @@
...
+    
+    def __repr__(self):
+        return f'Vector({self.x}, {self.y}, {self.z})'
+
+    def __add__(self, other):        
+        x = self.x + other.x
+        y = self.y + other.y
+        z = self.z + other.z
+        v = Vector(x, y, z)
+        return v
```
gikf, line 21:
```suggestion
        if not isinstance(other, Vector):
            return NotImplemented
        x = self.x + other.x
        y = self.y + other.y
        z = self.z + other.z
        return Vector(x, y, z)
```
`Vector` class doesn't know how to add itself to the `other`'s class, however the class of `other` might know how to add itself to `Vector`. The `return NotImplemented` will allow python to automatically try that possibility.

Applies to other methods too.

scientific-computing-with-python/learn-special-methods-by-building-a-vector-space/vectors.py
```diff
@@ -0,0 +1,87 @@
...
+    
+    def __mul__(self, other):
+        # Scalar multiplication
+        if isinstance(other, int) or isinstance(other, float):
+            self.x *= other
+            self.y *= other
+            self.z *= other
+            self.norm *= other
+            return self
```
gikf, line 37: I'd consider returning completely new `Vector` instance with multiplied attributes, instead of modifying self. Similarly like in other methods.

scientific-computing-with-python/learn-special-methods-by-building-a-vector-space/vectors.py
```diff
@@ -0,0 +1,96 @@
...
+    def __repr__(self):
+        return f'Vector({self.x}, {self.y}, {self.z})'
+
+    def __add__(self, other):
+        if not isinstance(other, Vector):
+            return NotImplemented
+        x = self.x + other.x
+        y = self.y + other.y
+        z = self.z + other.z        
+        return Vector(x, y, z)
```
gikf, line 22: I remembered another detail. It might be better to return the new instance of the _current_ class, instead of directly `Vector`. The difference is that if somebody uses `Vector` as a parent class, without changing method returning `Vector`, the child class will keep returning `Vector`, instead of the child class.

Therefore to make it more flexible, in case somebody uses `Vector` as a parent class:
```suggestion
        return self.__class__(x, y, z)
```

scientific-computing-with-python/learn-special-methods-by-building-a-vector-space/vectors.py
```diff
@@ -0,0 +1,96 @@
...
+        y = self.z * other.x - self.x * other.z
+        z = self.x * other.y - self.y * other.x
+        return Vector(x, y, z)
+    
+    def __eq__(self, other):
+        if not isinstance(other, Vector):
+            return NotImplemented
+        if self.x == other.x \
+        and self.y == other.y \
+        and self.z == other.z:
```
gikf, line 62:
```suggestion
        if self.x == other.x and self.y == other.y and self.z == other.z:
```
Nitpicking, it's not that long to split it on more lines.

zairahira:
The project is looking great 👍️

scientific-computing-with-python/learn-special-methods-by-building-a-vector-space/vectors.py
```diff
@@ -0,0 +1,97 @@
...
+        # Scalar product
+        elif isinstance(other, Vector):
+            x = self.x * other.x
+            y = self.y * other.y
+            z = self.z * other.z
+            scalar = (x**2 + y**2 + z**2)**0.5
+            return scalar
+        else:
+            other_type = other.__class__.__name__
+            raise TypeError(f"Multiplication is not defined between 'Vector' and '{other_type}'")
```
gikf, line 49:
```suggestion
        other_type = other.__class__.__name__
        raise TypeError(f"Multiplication is not defined between 'Vector' and '{other_type}'")
```
Because there's `return` for other cases this doesn't need to be in the `else`, indented.

On the side note - is there specific reason why class raises here exception on it own, instead of letting handle it to Python, like in other cases? Or just trying to show campers various approaches?

scientific-computing-with-python/learn-special-methods-by-building-a-vector-space/vectors.py
```diff
@@ -0,0 +1,97 @@
...
+        
+    def cross(self, other):
+        # Cross product
+        if not isinstance(other, Vector):
+            raise TypeError("Argument must be a 'Vector' instance")
```
gikf, line 54: Similar question here.

gikf (commented):
Nitpicking!

scientific-computing-with-python/learn-special-methods-by-building-a-vector-space/vectors.py
```diff
@@ -0,0 +1,97 @@
...
+        self.x = x
+        self.y = y
+        self.z = z
+        self.norm = (x**2 + y**2 + z**2)**0.5
+
+    def __str__(self):
+        return str((self.x, self.y, self.z))
+    
+    def __repr__(self):
+        return f'Vector({self.x}, {self.y}, {self.z})'
```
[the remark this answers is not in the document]
Dario-DC: Is it really necessary to replace `Vector` with `self.__class__.__name__`? The only case I can think of is when another class inherits from Vector but I'm not sure about which kind of class could be.

Dario-DC:
I've just refactored the code to include inheritance. Not sure if it's a good idea but in this way we can use this project to introduce inheritance and teach abstract base classes in the following project (equation solver project, which I'm working on)

ilenia-magoni (changes requested):
I really like this but it has some issues

scientific-computing-with-python/learn-special-methods-by-building-a-vector-space/vectors.py
```diff
@@ -0,0 +1,94 @@
...
+            return self.__class__(**kwargs)
+    
+        # Scalar product
+        elif isinstance(other, self.__class__):
+            args = [self.__getattribute__(i) * other.__getattribute__(i) for i in self.__dict__]
+            return (sum(arg**2 for arg in args))**0.5
+        
+        raise TypeError(f"Multiplication is not defined between '{self.__class__.__name__}' and '{other.__class__.__name__}'")
+    
+    def __eq__(self, other):
```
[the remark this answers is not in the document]
Dario-DC: I turned the `return NotImplemented` into a `raise TypeError`. Probably it's not ideal but it should work now.

ilenia-magoni (approved):

zairahira (approved):

PR #372, opened 2024-02-13
feat: add equation solver prototype

Dario-DC:
Checklist:

A possible project to add to the new Python curriculum.

### DIFF
--- a/scientific-computing-with-python/learn-inheritance-by-building-an-equation-solver/quadratic_equation_solver.py
+++ b/scientific-computing-with-python/learn-inheritance-by-building-an-equation-solver/quadratic_equation_solver.py
@@ -124,7 +124,7 @@ class Solver:
         self.equation.solve()
         self.equation.analyze()
     
-    def create_output(self):
+    def print_output(self):
         self._solve()
         output_string = '\n{:-^24}'.format(self.equation.type) + f'\n\n{str(self.equation):^24}\n\n'        
         output_string += '{:-^24}'.format('Solutions')
@@ -140,7 +140,6 @@ class Solver:
         print(output_string)
         
 
-eq1 = QuadraticEquation(-2, -3, 2)
-eq2 = LinearEquation(2, 4)
-s = Solver(eq2)
-s.create_output()
+eq = QuadraticEquation(-2, -3, 2)
+s = Solver(eq)
+s.print_output()
--- a/scientific-computing-with-python/learn-inheritance-by-building-an-equation-solver/quadratic_equation_solver.py
+++ b/scientific-computing-with-python/learn-inheritance-by-building-an-equation-solver/quadratic_equation_solver.py
@@ -1,186 +1,146 @@
+from abc import ABC, abstractmethod
 import re
 
-class QuadraticEquation: 
-    type = 'Equation'
 
-    def __init__(self, a=0, b=0, c=0):
-        if a == 0:
-            raise ValueError(f"'{self.__class__.__name__}.a' must be different from zero")
-        self.a = a
-        self.b = b
-        self.c = c        
-        self.line = self.__prettify_line() + ' = 0'
-        self.delta = self.b**2 - 4 * self.a * self.c
-        self.results = set()
-        self.details = set()
+class Equation(ABC):
+    """
+    Equation objects are instantiated providing the equation coefficients as arguments.
+    The coefficient order goes from the zero-th grade of x to the n-th grade.
+    Coefficients with the value of zero must be specified.
+    """
+    def __init__(self, *args):
+        if (self.grade + 1)!= len(args):
+            raise TypeError(f"'Equation' object takes {self.grade + 1} positional arguments but {len(args)} were given")
+        for arg in args:            
+            if not isinstance(arg, (int, float)):
+                raise TypeError("Coefficients must be of type 'int' or 'float'")
+        if args[-1] == 0:
+            raise ValueError('Highest order coefficient must be different from zero')
+        self.coefficients = {n: arg for n, arg in enumerate(args)}
+        self.results = []
+        self.details = []        
 
-    def __prettify_line(self):
-        line = ''
-        if self.a != 0:
-            if self.a == 1:
-                line += 'x² '
-            elif self.a == -1:
-                line += '-x² '
-            else:
-                line += f'{self.a}x² '
-        if self.b != 0:            
-           line += '{:+}x '.format(self.b)
-        if self.c != 0:            
-            line += '{:+}'.format(self.c)
-        line = line.replace('+1x', '+x').replace('-1x', '-x')
-        return line
-    
-    def create_output(self):
-        output_string = '\n{:-^24}'.format(self.type) + f'\n\n{self.line:^24}\n\n'
-        if self.results:
-            output_string += '{:-^24}'.format('Solutions')
-            output_string += '\n\n'
-            for result in self.results:
-                output_string += f'{result:^24}\n'
-            output_string += '\n'
-        if self.details:
-            output_string += '{:-^24}'.format('Details')
-            output_string += '\n\n' 
-            for detail in self.details:
-                output_string += detail
-            output_string += '\n'
-        return output_string
-    
     def __str__(self):
-        return self.create_output()
-    
-    @staticmethod    
-    def parse(expression, coeff={'a': 0, 'b': 0, 'c': 0}):
-        transformed_expr = QuadraticEquation.__validate(expression)
-        terms = re.split(r'(?<!^)(?=[\+-])', transformed_expr)        
-        for term in terms:
-            try:
-                spaceless_term = re.sub(r'\s', '', term)
-                if 'x**2' in spaceless_term:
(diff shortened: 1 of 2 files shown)


scientific-computing-with-python/learn-inheritance-by-building-an-equation-solver/quadratic_equation_solver.py
```diff
@@ -0,0 +1,186 @@
...
+        x_max = max(x1, x2)
+        e = '=' if '=' in self.sign else ''
+        if '>' in self.sign:            
+            self.results.update((f'x <{e} {round(x_min, 3)}', f'x >{e} {round(x_max, 3)}'))
+        elif '<' in self.sign:            
+            self.results = {f'{round(x_min, 3)} <{e} x <{e} {round(x_max, 3)}'}        
+        return self.results
+
+    def analyze(self):
+        raise AttributeError("'QuadraticInequality' object has no attribute 'analyze'")
```
gikf, line 174: To be honest there isn't that much inheritance between these two classes. Line above is a good example for that. One of the _SOLID_ principle is that it should be possible to replace the instance of parent with instance of child, without changing behavior. In other words child class should be doing everything (and possibly more) as the parent class in the same way. `analyze` would yell at somebody expecting that :slightly_smiling_face: 

It's not easy to come up with good inheritance example, but I think the the equations idea can still be used, with slight reorganizing and tackling it from different angle. I'd propose to introduce abstract base class `Equation` (I guess it doesn't necessarily needs to be abstract, but this is a great place for it) which will be a parent class of equations.

```python
from abc import ABC, abstractmethod
class Equation(ABC):
    """"""
    @abstractmethod
    def solve(self):
        pass

    @abstractmethod
    def details(self):
        pass

    @abstractmethod
    def coefficients(self):
        pass
```
Because of being abstract parent, it will make sure the child classes are implementing methods that are required for class to work with the _Solver_:

```python
class Solver:
    def __init__(self, equation):
        self.equation = equation
    def create_output(self):
        ...
    # other methods presenting/preparing result
```
`Solver` isn't a great name for this class, because the `equation` effectively solves itself on it own. I don't have anything better for it right now.

Then we have concrete equations, that must implement abstract methods from `Equation`. They can of course add more methods that apply to the equation type that they are, but the _core interface_ with which `Solver` uses them is defined in the abstract base class `Equation`:
```python
class QuadraticEquation(Equation):
    ....

class QuadraticInequality(Equation):
    ....
```
But also there can be simpler (or harder) concrete equations defined, if somebody (camper) wants and is able to create concrete class, which still will work with the _Solver_.
```python
class LinearEquation(Equation):
    ....
```

Dario-DC: I agree, I was struggling to find a good way to teach something useful and meaningful at the same time.

Probably, this is the right place to introduce abstract classes and it can improve the overall project. I'll try to restructure the prototype.

scientific-computing-with-python/learn-inheritance-by-building-an-equation-solver/quadratic_equation_solver.py
```diff
@@ -0,0 +1,146 @@
...
+    def __init__(self, equation):
+        if not isinstance(equation, Equation):
+            raise TypeError("Argument must be an Equation object")
+        self.equation = equation
+
+    def _solve(self):
+        self.equation.solve()
+        self.equation.analyze()
+    
+    def create_output(self):
```
ilenia-magoni, line 127:
```suggestion
    def print_output(self):
```
I would call this like this as it prints, but I think it's fine otherwise

scientific-computing-with-python/learn-inheritance-by-building-an-equation-solver/quadratic_equation_solver.py
```diff
@@ -0,0 +1,145 @@
...
+            if coefficient:
+                if n == 0:
+                    terms.append(f'{coefficient:+}')
+                elif n == 1:
+                    terms.insert(0, f'{coefficient:+}x')
+                else:
+                    terms.insert(0, f'{coefficient:+}x**{n}')
+        equation_string = ' '.join(terms)
+        equation_string += ' = 0'
+        return re.sub(r'(?<!\d)1(?=x)', '', equation_string.strip('+'))
```
zairahira, line 35: To improve readability, I would suggest to refactor the `__str__` method.

```suggestion
   def __str__(self):
        return self.format_equation()

    def format_equation(self):
        terms = []
        for n, coefficient in self.coefficients.items():
            if coefficient:
                if n == 0:
                    terms.append(f'{coefficient:+}')
                elif n == 1:
                    terms.insert(0, f'{coefficient:+}x')
                else:
                    terms.insert(0, f'{coefficient:+}x**{n}')
        equation_string = ' '.join(terms)
        equation_string += ' = 0'
        return re.sub(r'(?<!\d)1(?=x)', '', equation_string.strip('+'))
```

scientific-computing-with-python/learn-inheritance-by-building-an-equation-solver/quadratic_equation_solver.py
```diff
@@ -0,0 +1,145 @@
...
+        output_string += '\n'        
+        output_string += '{:-^24}'.format('Details')
+        output_string += '\n\n' 
+        for detail in self.equation.details:
+            output_string += detail
+        output_string += '\n'
+        print(output_string)
+        
+
+eq = QuadraticEquation(-2, -3, 2)
```
zairahira, line 143: The general formula for a quadratic equation is `ax**2 + bx + c = 0`, the coefficients should also be mapped in this sequence.

For example, I would interpret `eq = QuadraticEquation(-2, -3, 9)` as `-2x**2 - 3x + 9` instead of `9x**2 - 3x - 2`.

Currently it maps to `c+bx+ax**2`. It's not wrong, just different from the standard equation.

zairahira (commented):
I left a few suggestions, implementation is not compulsory though.

Rest looks good to me, I think this would be a good practice project to teach OOP concepts and the mathematical perspective of SCWP.

zairahira (approved):

scientific-computing-with-python/learn-inheritance-by-building-an-equation-solver/quadratic_equation_solver.py
```diff
@@ -0,0 +1,136 @@
...
+            concavity = "upwards"
+            min_max = "min"
+        else:
+            concavity = "downwards"
+            min_max = "max"
+        return {"x": x, "y": y, "min_max": min_max, "concavity": concavity}
+
+
+def solver(equation):
+    width: int
```
Dario-DC, line 103: Remember to remove unused `width` variable

scientific-computing-with-python/learn-inheritance-by-building-an-equation-solver/quadratic_equation_solver.py
```diff
@@ -0,0 +1,136 @@
...
+        output_string += f"{result:^24}\n"
+    output_string += f'\n{"Details":-^24}\n\n'
+    details = equation.analyze()
+    if isinstance(equation, LinearEquation):
+        slope, intercept = details.values()
+        details_list = [f"slope = {slope:>16.3f}", f"y-intercept = {intercept:>10.3f}"]
+    elif isinstance(equation, QuadraticEquation):
+        x, y, min_max, concavity = details.values()
+        coord = f"({x:.3f}, {y:.3f})"
+        details_list = [f"concavity = {concavity:>12}", f"{min_max} = {coord:>18}"]
```
Dario-DC, line 127:
```suggestion
    match details:
        case {'slope': slope, 'intercept': intercept}:
            details_list = [f"slope = {slope:>16.3f}", f"y-intercept = {intercept:>10.3f}"]
        case {"x": x, "y": y, "min_max": min_max, "concavity": concavity}:
            coord = f"({x:.3f}, {y:.3f})"
            details_list = [f"concavity = {concavity:>12}", f"{min_max} = {coord:>18}"]
```
I got to this point of creating the steps. We could keep using the structural pattern matching.

scientific-computing-with-python/learn-inheritance-by-building-an-equation-solver/quadratic_equation_solver.py
```diff
@@ -0,0 +1,136 @@
...
+    output_string += f"\n\n{equation!s:^24}\n\n"
+    output_string += f'{"Solutions":-^24}\n\n'
+    results = equation.solve()
+    match len(results):
+        case 0:
+            result_list = ["No real roots"]
+        case 1:
+            result_list = [f"x = {results[0]:.3f}"]
+        case 2:
+            result_list = [f"x1 = {results[0]:+.3f}", f"x2 = {results[1]:+.3f}"]
```
Dario-DC, line 116:
```suggestion
    match results:
        case []:
            result_list = ["No real roots"]
        case [x]:
            result_list = [f"x = {x:+.3f}"]
        case [x1, x2]:
            result_list = [f"x1 = {x1:+.3f}", f"x2 = {x2:+.3f}"]
```

PR #376, opened 2024-05-13
feat(curriculum): Added the coffee machine project

fhsinchy:
Checklist:

Added the coffee machine project to teach OOP in the Python curriculum.


scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
```diff
@@ -0,0 +1,153 @@
...
+        self.resources = {
+            "water": 300,
+            "milk": 200,
+            "beans": 100,
+        }
+
+    def report(self):
+        print(f"Water: {self.resources['water']}ml")
+        print(f"Milk: {self.resources['milk']}ml")
+        print(f"Coffee: {self.resources['beans']}g")
```
Dario-DC, line 21:
```suggestion
        print(f"Water: {self.resources['water']} ml")
        print(f"Milk: {self.resources['milk']} ml")
        print(f"Coffee: {self.resources['beans']} g")
```
Adding some space to the output

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+class Stash:
+    CURRENCY = '$'
+
+    def __init__(self):
+        self.profit = 0
+
+    def report(self):
+        print(f"Money: {self.CURRENCY}{self.profit}")
```
Dario-DC, line 96:
```suggestion
        print(f"Money: {self.CURRENCY} {self.profit}")
```

Dario-DC (commented):
Just few suggestions after a quick review

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
```diff
@@ -0,0 +1,153 @@
...
+class MenuItem:
+
+    def __init__(self, name, water, milk, beans, cost):
+        self.name = name
+        self.cost = cost
+        self.ingredients = {
+            'water': water,
+            'milk': milk,
+            'beans': beans,
+        }
```
gikf, line 52: I think it would be better to pass ingredients as a list/dictionary, without specifically requiring all of them (ie. espresso already needs `0` water, so including it in ingredients is a bit obsolete).

This could be done in two ways:
```suggestion
    def __init__(self, name, ingredients, cost):
        self.name = name
        self.cost = cost
        self.ingredients = ingredients
```
Alternatively, with **kwargs:
```suggestion
    def __init__(self, name, cost, **ingredients):
        self.name = name
        self.cost = cost
        self.ingredients = ingredients
```

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
```diff
@@ -0,0 +1,153 @@
...
+
```
gikf, line 53: Some simple string representation?
```python
    def __str__(self):
        return self.name
```

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
```diff
@@ -0,0 +1,153 @@
...
+            MenuItem(name="latte", water=200, milk=150, beans=24, cost=2.5),
+            MenuItem(name="espresso", water=50, milk=0, beans=18, cost=1.5),
+            MenuItem(name="cappuccino", water=250, milk=50, beans=24, cost=3),
+        ]
+
+    def get_all_menu_items(self):
+        options = ""
+        for item in self.menu:
+            options += f"{item.name}/"
+        return options
```
gikf, line 75:
```suggestion
        return '/'.join(item.name for item in self.items)
```
Or, after adding `__str__` to `MenuItem`:
```suggestion
        return '/'.join(self.items)
```

- - -
Probably it also could be changed to `__str__` of the `Menu`.

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
```diff
@@ -0,0 +1,153 @@
...
+
+    def report(self):
+        print(f"Money: {self.CURRENCY}{self.profit}")
+
+    def make_payment(self, price):
+        money_received = float(input('How much money are you giving me? '))
+        if money_received >= price:
+            change = round(money_received - price, 2)
+            print(f"Here is {self.CURRENCY}{change} in change.")
+            self.profit += price
```
gikf, line 103: Just to point out, `change` is virtually infinite.

What about making machine that can accept only card payments, so no need to worry about change in the project?

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+class Menu:
+
+    def __init__(self):
+        self.menu = [
+            MenuItem(name="latte", water=200, milk=150, beans=24, cost=2.5),
+            MenuItem(name="espresso", water=50, milk=0, beans=18, cost=1.5),
+            MenuItem(name="cappuccino", water=250, milk=50, beans=24, cost=3),
+        ]
```
gikf, line 69: More convenient would be creating menu items elsewhere:
```suggestion
    def __init__(self, items):
        self.items = items
```

If having default items is very needed, this pattern could be used to get the best from two worlds, but I'd advice against it.
```suggestion
    def __init__(self, items=None):
        if items is None:
            self.items = [
                MenuItem(name="latte", water=200, milk=150, beans=24, cost=2.5),
                MenuItem(name="espresso", water=50, milk=0, beans=18, cost=1.5),
                MenuItem(name="cappuccino", water=250, milk=50, beans=24, cost=3),
            ]
        else:
            self.items = items
```

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
```diff
@@ -0,0 +1,153 @@
...
+        print("Sorry that item is not available.")
+
+
+"""
+The Stash class represents the stash. It keeps track of the money in the machine.
+"""
+
+
+class Stash:
+    CURRENCY = '$'
```
gikf, line 90: Having this as class variable limits the further usage of the class. It could be changed to instance variable (possibly with default value), or just dropped.

Nevertheless currency and `cost` in `MenuItem` don't have any actual relation.

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
```diff
@@ -0,0 +1,153 @@
+from time import sleep
+
+"""
+The Maker class represents the coffee maker itself. It keep tracks of the water, milk,
+and beans in the machine. Makes coffee and reports on the resources.
+"""
+
+
+class Maker:
```
gikf, line 9: Maybe some methods to replenish the resources?

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
```diff
@@ -0,0 +1,153 @@
...
+
+
+def welcome():
+    print('''
+          WELCOME TO FCC COFFEE MACHINE!
+
+          ------ MENU ------
+          Espresso ($1.50)
+          Latte ($2.50)
+          Cappuccino ($3.00)
```
gikf, line 117: This is something that should be printed from actual items in the menu.

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
gikf, line 1: Ideally only the main part of code should print anything on it own, or based on the results/values from the class instance's methods. Methods instead of printing anything may return string.

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
```diff
@@ -0,0 +1,153 @@
+from time import sleep
+
+"""
+The Maker class represents the coffee maker itself. It keep tracks of the water, milk,
+and beans in the machine. Makes coffee and reports on the resources.
```
gikf, line 5: Am I correct to assume there's the distinction between coffee maker and actual coffee machine?

gikf (commented):
There's couple things to look at to make usage of the classes easier. See specific comments below.

gikf:
Also how is this supposed to be used? Only as a script? It might be better to make it work both as a script and as a module (allow importing from elsewhere).

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
```diff
@@ -0,0 +1,153 @@
...
+        print('\033[31mError. Please choose an available option.\033[m')
+    else:
+        beverage = menu.find_menu_item(user_choice)  # Encapsulates the result
+        sufficient_resources = coffee_maker.is_resource_sufficient(
+            beverage)  # TrueFalse result
+        sufficient_money = money_stash.make_payment(beverage.cost)
+        if sufficient_resources and sufficient_money:
+            print('Thank you! Allow us to make your beverage now...')
+            coffee_maker.make_coffee(beverage)
+            sleep(5)
```
Dario-DC, line 153: Maybe this is not the best place to implement the structural pattern matching but, as discussed during the last meeting, it would be nice to show it instead of coding a menu that is very similar to what we have in the expense tracker.

```suggestion
    match user_choice:
        case 'off':
            print('Shutting down!')
            is_on = False
        case 'report':
            coffee_maker.report()
            money_stash.report()
        case user_choice if menu.find_menu_item(user_choice) is None:
            print('\033[31mError. Please choose an available option.\033[m')
        case _:
            beverage = menu.find_menu_item(user_choice)  # Encapsulates the result
            sufficient_resources = coffee_maker.is_resource_sufficient(beverage)  # TrueFalse result
            sufficient_money = money_stash.make_payment(beverage.cost)
            if sufficient_resources and sufficient_money:
                print('Thank you! Allow us to make your beverage now...')
                coffee_maker.make_coffee(beverage)
                sleep(5)
```

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
fhsinchy: I'll need to think a bit more about that. I agree with this being outdated though.

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
fhsinchy: This'd be very nice addition.

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
fhsinchy: Yeah why not but for now I think this will do. What do you say?

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
fhsinchy: What do you recommend in this situation?

fhsinchy:
@gikf and @Dario-DC I have implemented some of the suggestions and updated the code. Please let me know what you think. I still have to move all the print statement to the main part of the code will do that tonight. Looking forward to your comments.

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
```diff
@@ -0,0 +1,131 @@
...
+    print()
+    for available_item in menu.menu_items:
+        print(available_item)
+    print()
+    print('Type "report" to check available resource \nand "off" to log out from machine.')
+    print()
+
+    options = menu.get_all_menu_items()
+    user_choice = str(
+        input(f'What would you like?\nOptions ({options}): ')).strip().lower()
```
gikf, line 117:
```suggestion
    user_choice = input(f'What would you like?\nOptions ({options}): ').strip().lower()
```
`str` is obsolete, `input` always returns string.

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
```diff
@@ -0,0 +1,131 @@
...
+    '''
+    def __init__(self, profit: int):
+        self.profit = profit
+
+
+    def report(self):
+        print(f'Money: ${self.profit}')
+
+    def make_payment(self, price: float):
+        money_received = float(input('How much money are you giving me? '))
```
gikf, line 82: This needs `try/except` to ensure number is entered.

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
gikf: It's not a big deal, it's just JS curriculum has certification project with cash register. Of course similarly complex implementation here is not needed at all.

There's a choice:
- Making simplified cash/change handling
- Making card only handling

To some extent both are simplifications, with spots where things are not like in the real thing. However I think with second option it wouldn't be as visible.

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
gikf: I think how it is now - creating menu items elsewhere - is the way to go. 👍

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
```diff
@@ -0,0 +1,131 @@
...
+        if money_received >= price:
+            change = round(money_received - price, 2)
```
Dario-DC, line 84: Maybe instead of starting the program with `Stash(profit=0)`, the initial money in the machine could be something like `50`. Then we could have a maximum amount of money you can use to pay (instead of +∞), depending on if the money in the machine is greater than the change.

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
```diff
@@ -0,0 +1,131 @@
...
+    def __init__(self, resources: Dict [str, int]):
+        self.resources = resources
+
+    def report(self):
+        for resource, amount in self.resources.items():
+            print(f'{resource}: {amount} ml')
+
+    def is_resource_sufficient(self, menu_item):
+        for ingredient, amount in menu_item.ingredients.items():
+            if amount >= self.resources.get(ingredient, 0):
```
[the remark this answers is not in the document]
fhsinchy: If I put greater than, we'll get one less cup of coffee. I think this is fine.

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
fhsinchy: Yeah I forgot, my bad my bad.

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
```diff
@@ -0,0 +1,131 @@
...
+    The Maker class represents the coffee maker itself. It keeps track of the water, milk,
+    and beans in the machine. Makes coffee and reports on the resources.
+    '''
+
+    def __init__(self, resources: Dict [str, int]):
+        self.resources = resources
+
+    def report(self):
+        for resource, amount in self.resources.items():
+            print(f'{resource}: {amount} ml')
```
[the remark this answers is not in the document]
fhsinchy: Added a simple fix for this one. I could've done something more complex and appropriate but I think this may be enough for a project of this magnitude.

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
fhsinchy: I have added exception handling here as well as in the `find_menu_item` method. Previously in cases where the method returns `None` the print statement in the main section used to fail. Now the method throws an exception. I thought about doing something like creating a custom exception like "ItemNotFoundError" but went with a plain exception at the end. Please let me know your thoughts on that too.

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
fhsinchy: @gikf can you give me an idea of how the card based flow may work out please?

@Dario-DC yeah that can be done. Let me see.

Dario-DC (commented):
A question about type annotations: are you also going to include return annotations?

fhsinchy:
> A question about type annotations: are you also going to include return annotations?

I didn't because Python can infer the type from return value itself.

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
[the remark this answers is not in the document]
gikf: What do you mean? If required `amount` is greater than resource `ingredient`, coffee cannot be made. If they are equal, or resource is greater then coffee can be made.

scientific-computing-with-python/learn-object-oriented-programming-by-building-a-coffee-making-machine/CoffeeMachine.py
gikf: I think it could be something as simple as:
1. Display the price.
2. Ask to authorize card payment - ie. with `input`, value `'Y'` authorizes it, anything else cancels transaction.

gikf:
Python's default interpreter - CPython - do not care about annotations at all (as long as they don't break syntax). They can be incorrect, there won't be any complaints. Only IDEs or external tools might be trying to keep them correct, including inferring when that's possible.

Dario-DC:
> Python's default interpreter - CPython - do not care about annotations at all (as long as they don't break syntax). They can be incorrect, there won't be any complaints. Only IDEs or external tools might be trying to keep them correct, including inferring when that's possible.

Nevertheless we can check if they are correct with tests (if we want to introduce them). Of course, explaining that in a different environment this check should be done with some external tool.

fhsinchy:
I can remove them for now since we're using CPython on the platform.

Dario-DC (commented):
I am a bit confused about that:
image

image

fhsinchy:
@Dario-DC I've implemented the changes you've suggested.

fhsinchy:
> I am a bit confused about that: image
> 
> image

Not sure what's happening here. I've never come across this particular issue during my tests.

Dario-DC:
> Not sure what's happening here. I've never come across this particular issue during my tests.

It happened a couple of times and then it worked. Then it happened again and always with the same input :thinking:

fhsinchy:
> > Not sure what's happening here. I've never come across this particular issue during my tests.
> 
> It happened a couple of times and then it worked. Then it happened again and always with the same input 🤔

The amount of espresso I've ordered from the machine after your comment is enough to keep me awake for the rest of 2024 😄

PR #377, opened 2024-05-30
fix: vector space prototype

Dario-DC:
Checklist:

Closes #XXXXX

Following the suggestions on [#54140](https://github.com/freeCodeCamp/freeCodeCamp/pull/54140):
- `__dict__` has been substituted with `vars()`
- `__getattribute__()` has been substituted with `getattr()`
- the condition of several `if` statements that were using `isinstance` has been replaced because it wasn't preventing from performing operations between `R2Vector` and `R3Vector` instances.

### DIFF
--- a/scientific-computing-with-python/learn-special-methods-by-building-a-vector-space/vectors.py
+++ b/scientific-computing-with-python/learn-special-methods-by-building-a-vector-space/vectors.py
@@ -2,69 +2,78 @@ class R2Vector:
     def __init__(self, *, x, y):
         self.x = x
         self.y = y
-    
+
     def norm(self):
-        return sum(val**2 for val in self.__dict__.values())**0.5
+        return sum(val**2 for val in vars(self).values()) ** 0.5
 
     def __str__(self):
-        return str(tuple(self.__getattribute__(i) for i in self.__dict__))
+        return str(tuple(getattr(self, i) for i in vars(self)))
 
     def __repr__(self):
-        arg_list = [f'{key}={val}' for key, val in self.__dict__.items()]
+        arg_list = [f"{key}={val}" for key, val in vars(self).items()]
         args = ", ".join(arg_list)
-        return f'{self.__class__.__name__}({args})'
-    
+        return f"{self.__class__.__name__}({args})"
+
     def __add__(self, other):
-        if not isinstance(other, self.__class__):
-            raise TypeError(f"Addition is not defined between '{self.__class__.__name__}' and '{other.__class__.__name__}'")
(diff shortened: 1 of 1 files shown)


scientific-computing-with-python/learn-special-methods-by-building-a-vector-space/vectors.py
```diff
@@ -2,69 +2,64 @@ class R2Vector:
...
-        if not isinstance(other, self.__class__):
-            raise TypeError(f"Subtraction is not defined between '{self.__class__.__name__}' and '{other.__class__.__name__}'")
-        kwargs = {i: self.__getattribute__(i) - other.__getattribute__(i) for i in self.__dict__}
+        if self.__class__ != other.__class__:
+            return NotImplemented
+        kwargs = {i: getattr(self, i) - getattr(other, i) for i in vars(self)}
         return self.__class__(**kwargs)
-    
+
     def __mul__(self, other):
```
zairahira, line 29: To prevent multiplication of `bool`, `type()` can be used instead of `isinstance`:

```py
def __mul__(self, other):
        # Scalar multiplication
        if type(other) in {int, float}:
            kwargs = {i: getattr(self, i) * other for i in vars(self)}
            return self.__class__(**kwargs)
        # Dot product
        elif self.__class__ == other.__class__:
            args = [getattr(self, i) * getattr(other, i) for i in vars(self)]
            return sum(args)
        return NotImplemented
```

Dario-DC: Maybe I should use `type()` for the other `if`s, too.. It's cleaner.

gikf: If preventing of operations with `bool` (which is subclass of `int`) is really needed it could be singled out:
```python
    def __mul__(self, other):
        # Scalar multiplication
        if isinstance(other, bool):
            return NotImplemented
        elif isinstance(other, (int, float)):
            kwargs = {i: getattr(self, i) * other for i in vars(self)}
            return self.__class__(**kwargs)
        # Dot product
        elif isinstance(other, self.__class__):
            args = [getattr(self, i) * getattr(other, i) for i in vars(self)]
            return sum(args)
        return NotImplemented
```

I wouldn't say that's necessarily an issue. `bool` is just that way - `True == 1 and False == 0`, but `True is not 1 and False is not 0`. It can be used in various _seemingly_ invalid math operations. For example `sum((True, True, False)) == 2`

PR #387, opened 2024-06-28
chore: adding hotel feedback form workshop

### LINKED ISSUE #386: chore: add workshop for teaching forms in frontend cert
In the HTML forms and Tables module, there will be a workshop to teach forms.

jdwilkin4:
Checklist:

Closes #386

### DIFF
--- a/frontend-cert/html-projects/hotel-feedback-form/index.html
+++ b/frontend-cert/html-projects/hotel-feedback-form/index.html
@@ -1,0 +1,105 @@
+<!--NOTE: This will be a workshop(step based tutorial)-->
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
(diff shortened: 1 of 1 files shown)


moT01 (commented):
This looks good @jdwilkin4 👍 Here's some things I noticed:

I would give the action a URL. I played with this in our editor - when you click submit and there's no URL, nothing happens - when there is a URL, it gives an alert that says "Normally this form would be submitted! It works. This will be submitted to: URL."

I think we should add a `name` attribute to every input... "Only form elements with a `name` attribute will have their values passed when submitting a form." There's a few without a `name`.

Some of the labels are missing a `for` attribute. I'm not sure if we really need to add that to all of them but when a label is associated with in an input, "a screen reader will read out the label when the user is focused on the form input." So it's better for accessibility.

moT01 (approved):

PR #389, opened 2024-06-28
chore: adding student table workshop

### LINKED ISSUE #388: chore: add workshop for teaching tables in frontend cert
This will be the second workshop in the HTML forms and tables module in the frontend cert

jdwilkin4:
Checklist:

Closes #388

### DIFF
--- a/frontend-cert/html-projects/student-exams-table/index.html
+++ b/frontend-cert/html-projects/student-exams-table/index.html
@@ -1,0 +1,61 @@
+<!--NOTE: This will be a workshop(step based tutorial)-->
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Calculus Final Exams Table</title>
+  </head>
+  <body>
+    <table>
+      <caption>
+        Calculus Final Exam Grades
+      </caption>
+
+      <thead>
+        <tr>
+          <th>Last Name</th>
+          <th>First Name</th>
+          <th>Grade</th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr>
+          <td>Davis</td>
+          <td>Alex</td>
+          <td>54</td>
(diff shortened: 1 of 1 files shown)


moT01 (commented):
This looks good 👍 Here's some things I found that we could potentially add:

"Improve accessibility by adding `scope` attributes to the `<th>` elements to specify that they are headers for their respective columns" e.g:

```
<th scope="col">Last Name</th>
<th scope="col">First Name</th>
<th scope="col">Grade</th>
```

We could add some styling via the attributes...

```
<table border="1" cellspacing="0" cellpadding="8" width="50%">

<thead bgcolor="#ccc">
```

Screenshot 2024-07-01 at 3 04 24 PM

jdwilkin4:
I like how the styles look but I am on the fence about using inline styles because we will need to be super careful about messaging.
So we will probably need to add a note in the steps for adding inline styles that while inline styles are an option, usually it is better to use an external stylesheets for seperation of concerns reasons.

moT01:
We definitely don't have to add them - but I these are different from inline styles, they are actual table element attributes.

jdwilkin4:
It looks like these attributes are deprecated according to MDN
https://developer.mozilla.org/en-US/docs/Web/API/HTMLTableElement/border
https://developer.mozilla.org/en-US/docs/Web/API/HTMLTableElement/cellPadding
https://developer.mozilla.org/en-US/docs/Web/API/HTMLTableElement/cellSpacing

jdwilkin4:
we could add a style element and add some internal styles for the heading styles

moT01:
> these attributes are deprecated

Nice catch. Yea, if they're deprecated, we probably shouldn't use them.

> we could add a style element

So, style code that wouldn't be visible to the user? I'm not sure I really like that cause it could be confusing. If they go create a table elsewhere, they won't know why it doesn't look good.

moT01 (approved):
Going to approve this. Feel free to add more if you think it's appropriate @jdwilkin4.

PR #409, opened 2024-07-01
feat: add hmtl video lab

### LINKED ISSUE #382: chore: add HTML only video page project to frontend cert
The basic HTML section will have a few labs for campers to do.

In the previous section of the course, campers will have learned about working with audio and videos. So they will have learned how to work with iframes and embedding videos onto a page.

For this issue, we need a lab that covers the following basic HTML topics: 

- headings
- images(optional)
- paragraphs
- main element
- section elements 
- iframes for freeCodeCamp YouTube videos

## Note about length of project

This is still early on in the cert.
So this should still be a smaller scale project.
So we could probably just have them add maybe 3-4 videos on the page.

## Acceptance criteria

- [ ] create a new folder of the project in the `frontend-cert/html-projects` directory
- [ ] add your completed project in an `index.html` file in that new folder

Dario-DC:
Checklist:

Closes #382

I'm not sure about what we want to add here, but I think that we can ask campers to create a page that embeds video lectures from the upcoming frontend certification.


jdwilkin4:
I was thinking we could use videos from our youtube page
https://www.youtube.com/c/Freecodecamp

hundreds of options to choose from there

jdwilkin4:
> embeds video lectures from the upcoming frontend certification.

I am not sure how long it will take for the upcoming frontend cert videos to be made available.
So using fcc youtube videos will be a better choice because it won't block this project from being added to the superblock in the short term

moT01 (commented):
Looking good @Dario-DC 👍 

Couple minor typos.

Only other suggestion would be to give the iframes size properties cause they're kind of small:

`<iframe width="560" height="315" src="https://www.youtube.com/embed/7K6bCewZzJU?si=gCcenfsrI0mH_h0m"></iframe>`

But maybe that isn't really taught before this. Pretty simple things to figure out though.

jdwilkin4:
iframe sizes will be taught in the lecture videos so it is fine to include it here

bbsmooth:
I would add one thing for accessibility. The `iframe` element must have an accessible name to meet WCAG requirements. Usually this is done by adding the `title` attribute to the `iframe` element, which is what I would recommend here. In this case, because these are going to be videos, the `title` attribute would probably be the title of the video with the word "video" at the end.

moT01:
LGTM if this is ready @Dario-DC 👍 

Name suggestion: Video Compilation Page

Dario-DC:
> LGTM if this is ready @Dario-DC 👍
> 
> Name suggestion: Video Compilation Page

Done :+1:

moT01:
Just to clarify, I meant that we could call it `Video Compilation Page` on the main site. Then users can make it a freeCodeCamp videos page, or a bike riding videos page.

PR #415, opened 2024-07-03
chore: adding intro strings workshop

### LINKED ISSUE #414: chore: add intro strings workshop to frontend cert
This workshop will cover the following

- const
- assignment
- template literals
- str.length property
- str.indexOf method

jdwilkin4:
## Summary
This will be the intro strings workshop at the beginning of the the JavaScript strings module.

It covers the following:

- const
- assignment
- template literals
- str.length property
- str.indexOf method

Checklist:

Closes #414

### DIFF
--- a/frontend-cert/js-projects/js-teaching-bot/script.js
+++ b/frontend-cert/js-projects/js-teaching-bot/script.js
@@ -1,0 +1,86 @@
+// second workshop in JS section
+
+// review log statements and basic string usage
+console.log("Hi there!");
+
+// review variables and assignment
+const botName = "teacherBot";
+const occupation = "teacher";
+
+// teach string concatenation using the + operator
+// console.log("My name is " + botName);
+// console.log("I am an experienced " + occupation);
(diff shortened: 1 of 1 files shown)


moT01 (approved):
LGTM @jdwilkin4 🎉 

A couple potential adjustments. You could maybe put the concatenation into a variable. Not sure if that would help to understand it better. e.g. `const sentence = "My name is  + botName";` and `console.log(sentence);` Or maybe one use of each - one concatenation in the log and another in a variable. Could do the similar with the template literals.

You could add the length to the existing log. e.g:

Step 1: 
add `console.log(Here is an example of using the length property on the word ${topic}.)`

Step 2:
change to `console.log(Here is an example of using the length property on the word ${topic}: ${topic.length}.)`

Just kind of throwing ideas out there. I'm fine with it how it is.

jdwilkin4:
Yeah, we can put some of these into variables and just log the variable.
It will be easier to test that way to

PR #433, opened 2024-07-05
feat: add css colors lab

### LINKED ISSUE #395: chore: add css colors lab project to frontend cert
The CSS Colors module will have a few labs for campers to do.

The previous workshop will have been the colored markers project here
https://www.freecodecamp.org/learn/2022/responsive-web-design/learn-css-colors-by-building-a-set-of-colored-markers/step-94

The workshop covers hsl, rgb, hex codes and named colors.
This lab doesn't need to use all of those different color models.
Just pick one to have campers work with here.

## length of project

Even though we will have user stories for both HTML and CSS in this project, it should still be a smaller project then the colored markers.
So maybe 10-15 user stories?

## Acceptance criteria

- [x] create a new folder of the project in the `frontend-cert/css-projects` directory
- [x] add your completed project in an `index.html` file and `styles.css` file in that new folder
- [x] create a markdown file for your proposed user stories. You should call it `user-stories.md` and place it in your new project folder

ihechikara:
Checklist:

Closes #395

### DIFF
--- a/frontend-cert/css-projects/css-colors/index.html
+++ b/frontend-cert/css-projects/css-colors/index.html
@@ -9,11 +9,11 @@
 <body>
     <h1>CSS Colors</h1>
     <div class="color-grid">
-        <div class="color-box color1">#FF0000</div>
-        <div class="color-box color2">#33FF57</div>
-        <div class="color-box color3">orange</div>
-        <div class="color-box color5">rgba(255, 255, 0, 1)</div>
-        <div class="color-box color6">purple</div>
+        <div class="color-box color1"></div>
+        <div class="color-box color2"></div>
+        <div class="color-box color3"></div>
+        <div class="color-box color5"></div>
+        <div class="color-box color6"></div>
     </div>
 </body>
 </html>
--- a/frontend-cert/css-projects/css-colors/index.html
+++ b/frontend-cert/css-projects/css-colors/index.html
@@ -3,11 +3,11 @@
 <head>
     <meta charset="UTF-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <title>CSS Colors Lab</title>
+    <title>Colored Boxes</title>
     <link rel="stylesheet" href="styles.css">
 </head>
 <body>
-    <h1>CSS Colors</h1>
+    <h1>Colored Boxes</h1>
     <div class="color-grid">
         <div class="color-box color1"></div>
         <div class="color-box color2"></div>
--- a/frontend-cert/css-projects/css-colors/index.html
+++ b/frontend-cert/css-projects/css-colors/index.html
@@ -12,8 +12,8 @@
         <div class="color-box color1"></div>
         <div class="color-box color2"></div>
         <div class="color-box color3"></div>
+        <div class="color-box color4"></div>
         <div class="color-box color5"></div>
-        <div class="color-box color6"></div>
     </div>
 </body>
 </html>
--- a/frontend-cert/css-projects/css-colors/index.html
+++ b/frontend-cert/css-projects/css-colors/index.html
@@ -1,0 +1,19 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>CSS Colors Lab</title>
+    <link rel="stylesheet" href="styles.css">
+</head>
+<body>
+    <h1>CSS Colors</h1>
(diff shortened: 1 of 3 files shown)


frontend-cert/css-projects/css-colors/user-stories.md
```diff
@@ -0,0 +1,25 @@
+1. You should set the background color for `body` to `#f4f4f4`.
+
+2. You should have a `div` with a class of `color-grid`.
+
+3. You should have five `div` elements within `div` created above.
```
jdwilkin4, line 5: I would reference the class here

```suggestion
3. You should have five `div` elements within the `.color-grid` `div`.
```

jdwilkin4 (commented):
Left a small comment 👍

frontend-cert/css-projects/css-colors/user-stories.md
```diff
@@ -0,0 +1,25 @@
...
+
+4. The first `div` should have these classes: `color-box color1`.
```
moT01, line 7: This is looking good @ihechikara 👍 

Here's some slight adjustments to the user stories that make it a little more specific for some things, but also more open-ended for some of the colors:

You should set the background color for `body` to `#f4f4f4`.
You should have a `div` with a class of `color-grid` to hold all your color elements.
You should have five `div` elements within the `.color-grid` element.
The five `div` elements should each have a class of `color-box` and `color#`, where `#` is the number of the order of that `div`. For example, `color1` for the first `div`, and `color2` for the second, etc.
All of your `.color-box` elements should have a set `width` and `height` so they are visible on the page.
The `.color1` element should have a `background-color` that uses a predefined (word) color value.
The `.color2` element should have a `background-color` that uses a hexadecimal color value.
The `.color3` element should have a `background-color` that uses an RGB color value.
The `.color4` element should have a `background-color` that uses a HSL color value.
The `.color5` element should have a `background-color` set.

I changed it so that one of each type of color values is required, but not defined - so users can pick the colors, but they have to use all types.

Also note that if we change the user stories to be more open-ended like this, the prototype won't fulfill them anymore.

Let me know what you think @ihechikara.

ihechikara: Thanks for the suggestions, @moT01! I've updated the PR - let me know what you think

frontend-cert/css-projects/css-colors/index.html
```diff
@@ -0,0 +1,19 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>CSS Colors Lab</title>
```
jdwilkin4, line 6: I would call this something colored boxes

```suggestion
    <title>Colored Boxes</title>
```

frontend-cert/css-projects/css-colors/index.html
```diff
@@ -0,0 +1,19 @@
...
+    <link rel="stylesheet" href="styles.css">
+</head>
+<body>
+    <h1>CSS Colors</h1>
```
jdwilkin4, line 10:
```suggestion
    <h1>Colored Boxes</h1>
```

frontend-cert/css-projects/css-colors/styles.css
```diff
@@ -0,0 +1,50 @@
...
+.color-grid {
+    display: grid;
+    grid-template-columns: repeat(auto-fill, minmax(150px, 1fr));
+    gap: 10px;
+    width: 100%;
+    max-width: 800px;
+}
+
+.color-box {
+    display: flex;
```
Dario-DC, line 24: I'm afraid FlexBox has not been taught yet by this point.

Dario-DC: But now I see that it's not required :+1:

moT01 (changes requested):
Here's some more suggestions @ihechikara...

When I run your code, I don't see anything. The prototype is supposed to fulfill the user stories. Here's how you can do that:

Give each of the `.color#` class a background color described by the user stories. e.g. hex, hsl, etc. Give each `.color-box` a width and height so they're visible on the page.

ihechikara:
I've added them back, @moT01.

> Also note that if we change the user stories to be more open-ended like this, the prototype won't fulfill them anymore.

I removed them initially because in your last suggestion👆 you mentioned that having the user stories more open-ended meant that the prototype won't fulfil the stories. I may have misunderstood that. 

Let me know what you think!

moT01 (approved):
LGTM @ihechikara 🎉

PR #436, opened 2024-07-08
chore: add to-do list prototype 

### LINKED ISSUE #393: chore: add second basic css lab project to frontend cert
The Basic CSS module will have a few labs for campers to do.

The previous workshop will have been the cafe menu project here
https://www.freecodecamp.org/learn/2022/responsive-web-design/learn-basic-css-by-building-a-cafe-menu/step-91

## length of project

Even though we will have user stories for both HTML and CSS in this project, it should still be a smaller project then the cafe menu project.
So maybe 10-15 user stories?

## Acceptance criteria

zairahira:
Checklist:

Closes #393

### DIFF
--- a/frontend-cert/css-projects/todo-list-stylized/user-stories.md
+++ b/frontend-cert/css-projects/todo-list-stylized/user-stories.md
@@ -1,0 +1,21 @@
+1. You should have an `h1` element with the title `My To-Do List`.
+
+2. You should have an unordered list with the class `todo-list`.
+
+3. Inside the unordered list, you should have 4 list items. Each list item should have a checkbox, a task label and two links that point to a valid URL. 
+
+4. The first link in each list item should have the text `Edit` and the second link should have the text `Delete`. The links should also have the class `edit` and `delete` respectively.
+
+5. The list items should have a bottom margin of `15px` and a line height of `1.6`.
+
+6. Each list item should have squared bullets.
+
+7. The links should be `blue` initially and have a default of no underline. 
+
+8. When you hover over the links, the font color should change to `red`.
+
+9. Visited links should be `purple`.
+
+10. When a link is focused, there should be a `blue` border around it.
+
+11. When a link is being clicked, its color should change to `green`.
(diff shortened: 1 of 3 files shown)


frontend-cert/css-projects/todo-list-stylized/user-stories.md
```diff
@@ -0,0 +1,21 @@
...
+
+2. You should have an unordered list with the class `todo-list`.
+
+3. Inside the unordered list, you should have 4 list items. Each list item should have a checkbox, a task label and two links that point to a valid URL. 
+
+4. The first link in each list item should have the text `Edit` and the second link should have the text `Delete`. The links should also have the class `edit` and `delete` respectively.
+
+5. The list items should have a bottom margin of `15px` and a line height of `1.6`.
+
+6. Each list item should have squared bullets.
```
jdwilkin4, line 11: They haven't learned about list style type here, so we should remove this requirement

zairahira: In the project sheet, there is a lecture with the title "How to work with the different list-style properties" just above this lab.
I assumed that some list style types would be taught there.
Should I still remove the `list-style-type` from the prototype?

moT01 (commented):
Generally, I like this @zairahira 👍 - but here's a couple notes...

All the links should have `target=_blank` I think.

It's a little strange that edit and delete buttons link to external URLs. Can we find another way to include a link with those user stories? Like maybe make the link part of the todo list item? "All of your todo items should have a link in them" maybe? If we did that, we could remove the edit and delete buttons.

It also feels a little strange to have the checkbox and the squares on the left. I would be fine with that - just mentioning it.

zairahira:
@moT01, thanks for the feedback. I have made the following changes:

- Removed `delete` and `edit` buttons.
- Added a sub-item in the list item that should use a link.
- Removed the sqaure bullet for each main item.
- The sub-items should have a `list-style-type` of `square`.
- Links now have `target=_blank`.
- The links' `href` is currently `#`.

If these look fine, I'll update the user stories.

moT01:
Yea, I think that looks good @zairahira 👍 

I would suggest to format your HTML and there's some comments in the CSS you could remove.

moT01 (approved):

PR #445, opened 2024-07-10
feat: add css variables lab

### LINKED ISSUE #403: chore: add lab project for CSS variables in frontend cert
The CSS variables module will a lab for campers to do.

The previous workshop will have been the city skyline project here
https://www.freecodecamp.org/learn/2022/responsive-web-design/learn-css-variables-by-building-a-city-skyline/step-118

So you will need to create a small lab project focused on creating custom css properties for things colors.

## Acceptance criteria

Dario-DC:
Checklist:

Closes #403

It's pretty rough and the CSS needs improvements, but the main things are already laid out.

image

### DIFF
--- a/frontend-cert/css-projects/meeting-coordinator/index.html
+++ b/frontend-cert/css-projects/meeting-coordinator/index.html
@@ -4,79 +4,88 @@
 <head>
     <meta charset="utf-8">
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
-    <title>Meeting Coordinator</title>
+    <title>Meeting Scheduler</title>
     <link rel="stylesheet" href="styles.css">
 </head>
 
 <body>
     <main>
-        <div class="time" >
-            <span id="h9">9:00 AM</span>
-            <span id="h10">10:00 AM</span>
-            <span id="h11">11:00 AM</span>
-            <span id="h12">12:00 AM</span>
-            <span id="h1">1:00 PM</span>
-            <span id="h2">2:00 PM</span>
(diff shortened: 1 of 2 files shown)


jdwilkin4 (commented):
I feel like a table would be more semantic then divs and spans here.
Also, it looks like there are duplicate ids with range1 -8
Otherwise, it looks good so far 👍

moT01:
Looking good @Dario-DC 👍

I think I would just remove all the id's in the table. I'm not sure I see why they're needed - I think the classes alone are enough.

Name suggestion: `Availability Table` ?

You could add one header at the top so people can give it something unique. Maybe "Meeting Time" or "Training Session"

I don't know if users know linear gradients at this point. If not, you could maybe do HTML elements for the legend instead.

It might be nice to work another variable or two in there somewhere. Maybe with the text color, or margins or something.

User story suggestion:
You should have a table with at least 3 columns and 5 rows.

Dario-DC:
Duplicate ids are an oversight. I had a single column at the beginning and then I forgot to fix it.

I'm going to:
- add a heading
- convert the divs and spans into a table

They should know gradients at this point, the city skyline workshop comes before this lab.
But I used linear gradients in a previous lab #400 too. So, if they haven't been taught yet I should change the other lab.

moT01:
Sorry for the confusion. I meant the name for the project on the main site as "Availability Table".  You can keep "Meeting Scheduler" or whatever for the prototype.

moT01 (approved):
LGTM 👍

PR #448, opened 2024-07-10
fix: improve cookie recipe prototype

moT01:
Since we are hoping to have all the prototypes as a live demo users can click on, I thought this one could use some improvements. You can see it in action here: https://mot01.github.io/html-lab-1/

Really, I was creating a PR to the main repo to add this lab and wanted to include the link, so I gave it some improvements.

Also, I couldn't help myself - I changed the user stories slightly so users can create any recipe they want instead of the cookies. We can revert that if we want.

Checklist:

Closes #XXXXX


jdwilkin4 (approved):
This looks good to me.
For the demo project PR we should add the viewport meta element though.

Dario-DC (approved):

PR #457, opened 2024-07-12
feat: added objects workshop project

### LINKED ISSUE #424: chore: add objects workshop to frontend cert(JS only)
The frontend cert will have a module just for objects.
There will be two lecture blocks going into more detail about how to work with objects and their methods.
But we need a short(like 40 steps or less) workshop, for them to learn about objects interactively.

- let
- const
- assignment
- reassignment
- basic string usage
- console.log
- camelCase
- template literals
- str.length property
- str.indexOf method
- Number()
- Math.floor
- Math.random
- prompt()
- ===
- if/else if/else
- comparison operators(< >=)
- logical operators (|| &&)
- array basics
- push
- pop
- shift
- unshift

## Acceptance criteria

Ksound22:
Checklist:

Closes #424

### DIFF
--- a/frontend-cert/js-projects/student-grade-tracker/script.js
+++ b/frontend-cert/js-projects/student-grade-tracker/script.js
@@ -52,6 +52,15 @@ console.log(student1); //average score should not be null anymore
 // repeat for the remaining 2 students
 
 // Calculate grade with if/else
+/*
+Grading system
+90 - 100: A
+80 - 89: B
+70 - 79: C
+60 - 69: D
+50 - 59: E
+<50: F
+*/
 if (student1.averageScore >= 90 && student1.averageScore <= 100) {
   student1.grade = 'A';
 } else if (student1.averageScore >= 80 && student1.averageScore < 90) {
--- a/frontend-cert/js-projects/student-grade-tracker/script.js
+++ b/frontend-cert/js-projects/student-grade-tracker/script.js
@@ -1,5 +1,7 @@
+// start with an empty array - objects will be pushed into it later
 const students = [];
 
+// Have campers create 4 or more objects – but working with more than 3 might make thigns too repetitious and redundant
 const student1 = {
   name: 'John',
   scores: [88, 90, 85],
@@ -27,10 +29,12 @@ const student4 = {
   grade: '',
 };
 
+// introduce push here
 students.push(student1, student2, student3, student4);
 
 console.log(students);
 
+// introduce pop
 students.pop(); // remove the last student so three are two left to work with
 
 let student1TotalScore =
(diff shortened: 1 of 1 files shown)


frontend-cert/js-projects/student-grade-tracker/script.js
```diff
@@ -0,0 +1,107 @@
...
+
+console.log(student1TotalScore);
+// repeat for the two other students
+
+let student1AverageScore = student1TotalScore / student1.scores.length;
+console.log(student1AverageScore);
+
+// round the average scpre to two decmal places
+let student1RoundedAvg = Math.floor(student1AverageScore * 100) / 100;
+console.log(student1RoundedAvg);
```
jdwilkin4, line 51: In the numbers and math module, they will have learned about the `toFixed` method in the lectures. It looks like I left that out of the issue. 

But we should leave a note to allow for multiple correct answers when broken up into steps and tests.
Also, most campers will probably use this answer instead because they will have learned about it earlier.

```js
let student1RoundedAvg = (student1TotalScore / student1.scores.length).toFixed(2);
```
so we might consider using that in the final seed code instead

frontend-cert/js-projects/student-grade-tracker/script.js
```diff
@@ -0,0 +1,107 @@
...
+    student1.averageScore +
+    ' with the grade ' +
+    student1.grade +
+    '.'
+);
+
+// introduce template literals
+console.log(
+  `Student1, with the name ${student1.name} has an average score of ${student1.averageScore} with the grade ${student1.grade}.`
+);
```
jdwilkin4, line 103: I wouldn't have campers write this out twice. 

They will have already learned about both types of string concatenation.
I would just have one step where they create the string and them log it to the console.
That way it is up to them if they want to use template literals or the `+` operator for concatenating strings. 
Also, assigning the concatenated string to a variable will make it easier to test without relying on all regex tests.

So something like this instead 

```js
const student1GradeOutput = `Student1, with the name ${student1.name} has an average score of ${student1.averageScore} with the grade ${student1.grade}.`
console.log(student1GradeOutput)
```

frontend-cert/js-projects/student-grade-tracker/script.js
```diff
@@ -0,0 +1,107 @@
...
+
+// reoeat for the remaining 2 students
+
+console.log(students); // at this point, all the objects in the students array should have a value for all fields
```
jdwilkin4, line 107: At this point, I actually wouldn't have them repeat the if/else chain two more times because then we are teaching them to be repetitive. I would suggest removing this part.

Since functions are not taught yet, I would frame the narrative for the `if/else` statement to just say "Now we are going to get the student grade just for the first student. In a future module, you will learn about functions which will make this code even more resuable"

or something like that.

jdwilkin4 (changes requested):
I think this looks good.
Left a few comments 👍

jdwilkin4 (approved):

PR #467, opened 2024-07-14
feat: add styling forms lab 

### LINKED ISSUE #396: chore: add styling forms lab project to frontend cert
The Styling forms module will have a few labs for campers to do.

The previous workshop will have been the registration form project here
https://www.freecodecamp.org/learn/2022/responsive-web-design/learn-html-forms-by-building-a-registration-form/step-65

For this lab, we want to have a few user stories focused on adding styles to forms.

## length of project

Maybe 10-15 user stories?

## Acceptance criteria

ihechikara:
Checklist:

Closes #396

### DIFF
--- a/frontend-cert/css-projects/styling-forms/index.html
+++ b/frontend-cert/css-projects/styling-forms/index.html
@@ -1,0 +1,26 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>Styled Form Project</title>
+    <link rel="stylesheet" href="styles.css">
+</head>
+<body>
+    <div class="form-container">
+        <form>
+            <h2>Contact Us</h2>
+            <label for="name">Name:</label>
+            <input type="text" id="name" name="name" required>
+
+            <label for="email">Email:</label>
+            <input type="email" id="email" name="email" required>
+
+            <label for="message">Message:</label>
+            <textarea id="message" name="message" required></textarea>
+
+            <button type="submit">Submit</button>
+        </form>
+    </div>
+</body>
+</html>
--- a/frontend-cert/css-projects/styling-forms/styles.css
+++ b/frontend-cert/css-projects/styling-forms/styles.css
@@ -1,0 +1,53 @@
+body {
+    display: flex;
+    justify-content: center;
+    align-items: center;
+    height: 100vh;
+    margin: 0;
+    background: linear-gradient(45deg, #83a4d4, #b6fbff);
+    font-family: Arial, sans-serif;
+}
+
+.form-container {
+    background-color: white;
+    border-radius: 10px;
+    padding: 20px;
+    width: 300px;
+    text-align: center;
+}
+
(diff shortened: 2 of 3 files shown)


Dario-DC (commented):
It looks good :+1:  I left a few comments.

I have a doubt about using CSS property names with and without backticks. I'm okay with not surrounding property names for background color, border, or font size. But sometimes it seems weird to me. For example, border radius, or padding. 

I'm saying this, mainly because there's no accurate translation of those, and I think it might lead to inaccuracies if translated. Of course this may not be true for other languages.

ihechikara:
> It looks good 👍 I left a few comments.
> 
> I have a doubt about using CSS property names with and without backticks. I'm okay with not surrounding property names for background color, border, or font size. But sometimes it seems weird to me. For example, border radius, or padding.
> 
> I'm saying this, mainly because there's no accurate translation of those, and I think it might lead to inaccuracies if translated. Of course this may not be true for other languages.

@Dario-DC, I decided to put most of the property names within backticks. Based on your suggestion, I think that'll make translating them more straightforward. Let me know what you think.

moT01 (approved):
This looks good @ihechikara 👍 I would be fine it as-is, but here's some suggestions...

Make some of the text of the user stories more flexible - so users don't have to use exact values. For instance, instead of:

``A `margin-bottom` of `10px`.``, Maybe just `A margin on the bottom.` So people can use a value they want.

It's not necessary to change though, just mentioning it. You have several user stories that do allow for some flexible.

I also like `#4`: ``Each `input` and `textarea` element should have their corresponding `label` element with a `for` attribute set to the element `id`.`` I feel like this makes people think about what they have to do more, instead of just copying what they're told to do. For example one says to add ``A `padding` of `10px`.`` - It's fine to have some of these, but it feels like users don't really have to think about that. Again, just mention it - maybe something to keep in mind for future labs.

I would be fine getting this in if you are @ihechikara 👍  Feel free to merge it, or make adjustments if you think some of those should be changed.

ihechikara:
> This looks good @ihechikara 👍 I would be fine it as-is, but here's some suggestions...
> 
> Make some of the text of the user stories more flexible - so users don't have to use exact values. For instance, instead of:
> 
> `` A `margin-bottom` of `10px`. ``, Maybe just `A margin on the bottom.` So people can use a value they want.
> 
> It's not necessary to change though, just mentioning it. You have several user stories that do allow for some flexible.
> 
> I also like `#4`: `` Each `input` and `textarea` element should have their corresponding `label` element with a `for` attribute set to the element `id`. `` I feel like this makes people think about what they have to do more, instead of just copying what they're told to do. For example one says to add `` A `padding` of `10px`. `` - It's fine to have some of these, but it feels like users don't really have to think about that. Again, just mention it - maybe something to keep in mind for future labs.
> 
> I would be fine getting this in if you are @ihechikara 👍 Feel free to merge it, or make adjustments if you think some of those should be changed.

Thanks for the suggestions, @moT01! I've made the user stories more flexible. Just wondering - should I do the same with `#3` in the user stories? Let me know what you think.

moT01:
LGTM 👍

PR #488, opened 2024-07-21
feat: add second functions lab

### LINKED ISSUE #456: chore: add second functions lab to frontend cert (JS only)
Functions are a topic that confuse a lot of campers.
So we should have extra labs for this

The previous workshop and lecture videos will teach campers about functions.

Here is what they will know so far.

- let
- const
- assignment
- reassignment
- basic string usage
- console.log
- camelCase
- template literals
- str.length property
- str.indexOf method
- Number()
- Math.floor
- Math.random
- prompt()
- ===
- if/else if/else
- comparison operators(< >=)
- logical operators (|| &&)
- array basics
- push
- pop
- shift
- unshift
- basic object usage
- for loop
- for of loop
- do/while loop
- while loop
- forEach loop
- regular functions
- arrow functions

## Acceptance criteria

ihechikara:
Checklist:

Closes #456

### DIFF
--- a/frontend-cert/js-projects/second-functions/script.js
+++ b/frontend-cert/js-projects/second-functions/script.js
@@ -1,0 +1,32 @@
+function askQuestion(question) {
+    return prompt(question);
+}
+
+function checkAnswer(userAnswer, correctAnswer) {
+    return userAnswer.toLowerCase() === correctAnswer.toLowerCase();
+}
+
+const runQuiz = () => {
+    const questions = [
+        { question: "What is the capital of France?", answer: "Paris" },
+        { question: "What is 2 + 2?", answer: "4" },
+        { question: "What is the largest planet in our solar system?", answer: "Jupiter" },
+        { question: "What is the smallest prime number?", answer: "2" },
+    ];
+
+    let score = 0;
+
+    questions.forEach((q) => {
+        const userAnswer = askQuestion(q.question);
+        if (checkAnswer(userAnswer, q.answer)) {
+            console.log("Correct!");
+            score++;
+        } else {
+            console.log(`Wrong. The correct answer is ${q.answer}.`);
+        }
+    });
+
+    console.log(`You scored ${score} out of ${questions.length}.`);
+}
+
+runQuiz();
--- a/frontend-cert/js-projects/second-functions/script.js
+++ b/frontend-cert/js-projects/second-functions/script.js
@@ -1,9 +1,5 @@
-function askQuestion(question) {
-    return prompt(question);
-}
-
 function checkAnswer(userAnswer, correctAnswer) {
-    return userAnswer.toLowerCase() === correctAnswer.toLowerCase();
+    return userAnswer === correctAnswer;
 }
 
 const runQuiz = () => {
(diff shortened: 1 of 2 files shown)


frontend-cert/js-projects/second-functions/user-stories.md
```diff
@@ -0,0 +1,20 @@
+1. You should define a function named `askQuestion` with a `question` parameter. The function should return a prompt with `question` as its parameter.
+
+2. You should define a function named `checkAnswer` with two parameters: `userAnswer` and `correctAnswer`. The fucntion should return the comparison of `userAnswer.toLowerCase()` and `correctAnswer.toLowerCase()`.
```
zairahira, line 3:
```suggestion
2. You should define a function named `checkAnswer` with two parameters: `userAnswer` and `correctAnswer`. The function should return the comparison of `userAnswer.toLowerCase()` and `correctAnswer.toLowerCase()`.
```

frontend-cert/js-projects/second-functions/user-stories.md
```diff
@@ -0,0 +1,20 @@
...
+
+3. You should define an arrow function named `runQuiz`. Step 4-7 should be created within the `runQuiz` function.
+
+4. You should create an array named `questions`. The array should have objects with `question` and `answer` properties. For example: `{ question: "What is the capital of France?", answer: "Paris" }`, `{ question: "What is 2 + 2?", answer: "4" },`, and so on. Create as many objects as you want.
```
zairahira, line 7: I think here we should put a limit and provide those questions so it is easier to test them.
What do you think about that?

ihechikara: I agree. I've updated the user stories 👍

frontend-cert/js-projects/second-functions/script.js
```diff
@@ -0,0 +1,32 @@
+function askQuestion(question) {
+    return prompt(question);
+}
+
+function checkAnswer(userAnswer, correctAnswer) {
+    return userAnswer.toLowerCase() === correctAnswer.toLowerCase();
```
moT01, line 6: Is `toLowerCase` covered before this?

frontend-cert/js-projects/second-functions/script.js
```diff
@@ -0,0 +1,32 @@
+function askQuestion(question) {
+    return prompt(question);
+}
```
moT01, line 3: I wonder if we should get rid of this - you could just put `prompt(question)` in the `forEach`? That would simplify things a little I think. The user stories would need to be adjusted slightly.

frontend-cert/js-projects/second-functions/user-stories.md
```diff
@@ -0,0 +1,20 @@
+1. You should define a function named `askQuestion` with a `question` parameter. The function should return a prompt with `question` as its parameter.
+
+2. You should define a function named `checkAnswer` with two parameters: `userAnswer` and `correctAnswer`. The function should return the comparison of `userAnswer.toLowerCase()` and `correctAnswer.toLowerCase()`.
+
+3. You should define an arrow function named `runQuiz`. Step 4-7 should be created within the `runQuiz` function.
+
+4. You should create an array named `questions`. The array should have objects with `question` and `answer` properties. For example, one object should look like this: `{ question: "What is 2 + 2?", answer: "4" }`. You should create four objects.
```
moT01, line 7:
```suggestion
4. You should create an array named `questions`. The array should have objects with `question` and `answer` properties. For example, one object should look like this: `{ question: "What is 2 + 2?", answer: "4" }`. You should create at least four objects.
```

This gives a little more flexibility in user projects.

moT01 (commented):
This is looking good @ihechikara 👍 I left a few suggestions. 

Additionally, I wonder if we should adjust the logs a little bit. When I run the code and enter four correct answers, the logs just say `Correct!` four times. Maybe something to make the logs look like:

```
Question: What is the capital of France?
Correct!

Question: What is 2 + 2?
Wrong. The correct answer is 4
```

That's not really needed, I just think it would be a little better.

moT01 (approved):
LGTM 🎉

PR #490, opened 2024-07-23
chore: add n-th Fibonacci number lab

### LINKED ISSUE #474: chore: add Nth fibonacci number dynamic programming lab to frontend cert(JS only)
There will be lectures videos to talk about dynamic programming.
But we need a lab for them to do it themselves.

A good problem would be to have a lab finding the Nth fibonacci number
https://www.geeksforgeeks.org/program-for-nth-fibonacci-number/

## Acceptance criteria

Dario-DC:
Checklist:

Closes #474

### DIFF
--- a/frontend-cert/js-projects/nth-fibonacci-number/user-stories.md
+++ b/frontend-cert/js-projects/nth-fibonacci-number/user-stories.md
@@ -1,3 +1,5 @@
+1. You should define an empty array named `results` in the global scope.
 1. You should create a function named `fibonacci`.
-1. The `fibonacci` function should take single parameter, which is an integer number.
-1. A call to `fibonacci(n)` should return the `n`-th number from the Fibonacci sequence, in which each number is the sum of the two preceding numbers.
+1. The `fibonacci` function should take single parameter, which is a positive integer number.
+1. A call to `fibonacci(n)` should use a dynamic programming approach to compute and return the `n`-th number from the Fibonacci sequence, in which each number is the sum of the two preceding numbers.
+1. Each computed number at the position `n` in the Fibonacci sequence should be stored at the index `n` of the `results` array.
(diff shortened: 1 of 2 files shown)


zairahira:
It seems like the code is following a recursive approach rather than a dynamic approach. The goal of this lab isn't recursion according to the issue description.

Dario-DC:
I refactored the code to use a bottom-up approach. This should count as dynamic @zairahira 

I'm asking to create an array in the global scope to store the computed sequence numbers because I thought it would be easier to test. I'm not sure, though.

frontend-cert/js-projects/nth-fibonacci-number/user-stories.md
```diff
@@ -0,0 +1,5 @@
+1. You should define an empty array named `results` in the global scope.
+1. You should create a function named `fibonacci`.
+1. The `fibonacci` function should take single parameter, which is a positive integer number.
```
zairahira, line 3:
```suggestion
1. The `fibonacci` function should take a single parameter, which is a positive integer.
```

zairahira:
This version implements the dynamic approach and it's looking good 🎉

zairahira (approved):

moT01 (commented):
Pretty minor, but I would change the results variable to `sequence`.

I don't think it's working quite right, if I log `fibonacci(1)` I should get `0`, the first number in the sequence, but I get `1`. Here's some adjusted code:

```
const results = [0,1];

const fibonacci = (n) => {
    if (n <= 2) return results[n-1];

    for (let i = 2; i <= n; i++) {
        results.push(results[i - 1] + results[i - 2]);
    }
    return results[n-1]
}

console.log(fibonacci(10));
```

You don't have to use that, just an example.

Dario-DC:
Good catch, I didn't realized it @moT01

moT01:
The prototype doesn't fulfill the user stories anymore. "You should define an empty array named `sequence` in the global scope." We should either adjust the user story, or write a different solution that starts with an empty array.

Dario-DC:
> The prototype doesn't fulfill the user stories anymore. "You should define an empty array named `sequence` in the global scope." We should either adjust the user story, or write a different solution that starts with an empty array.

My bad, I updated the name and forgot to fix the rest accordingly.

moT01 (approved):

PR #491, opened 2024-07-23
feat: add inventory management lab prototype

### LINKED ISSUE #434: chore: add lab for JS fundamentals (JS only)
There is a whole module just for reviewing the JS fundamentals before they moving onto working with the DOM and other concepts.
Campers will have just finished the pyramid generator project here 
https://www.freecodecamp.org/learn/javascript-algorithms-and-data-structures-v8/learn-introductory-javascript-by-building-a-pyramid-generator/step-118

So we need a lab for them to do immediately after that.
Here is what they will know so far

- let
- const
- assignment
- reassignment
- basic string usage
- console.log
- camelCase
- template literals
- str.length property
- str.indexOf method
- Number()
- Math.floor
- Math.random
- prompt()
- ===
- if/else if/else
- comparison operators(< >=)
- logical operators (|| &&)
- array basics
- push
- pop
- shift
- unshift
- basic object usage
- for loop
- for of loop
- do/while loop
- while loop
- forEach loop
- regular functions
- arrow functions

## Acceptance criteria

Dario-DC:
Checklist:

Closes #434


frontend-cert/js-projects/inventory-management-app/script.js
```diff
@@ -0,0 +1,67 @@
+let inventory = [];
+
+const findProductIndex = (productName) => {
```
moT01, line 3: It doesn't really matter how users find the index, but I'm curious if they will know the `indexOf` method at this point?

frontend-cert/js-projects/inventory-management-app/user-stories.md
```diff
@@ -0,0 +1,4 @@
...
+1. You should create a function named `addProduct` that takes a product object as its argument. If the product is already present in the inventory, the function should update its `quantity` value and log to the console the product name followed by a space and `quantity updated`. Otherwise, the function should push the product to the `inventory` array and log to the console the product name followed by a space and `added to inventory`. 
+1. You should create a function named `removeProduct` that takes the product name and quantity as its arguments. The function should subtract the passed quantity from the corresponding product object inside the inventory and log the string `Remaining <product-name> pieces: <product-quantity>` to the console, where `<product-name>` should be replaced by the product name, and `<product-quantity>` should be replaced by the product quantity. If the quantity after the subtraction is zero, the product object should be removed from the inventory. Instead, if the quantity in the inventory is not enough to perform the subtraction, the function should log the string `Not enough <product-name> available, remaining pieces: <product-quantity>` to the console.
```
moT01, line 4: I feel like these two could be split up into several user stories.

moT01 (approved):
LGTM @Dario-DC 👍 I left a few suggestions, feel free to make adjustments if you want, or just merge it if you're happy.

frontend-cert/js-projects/inventory-management-app/script.js
Dario-DC: I guess they should but I didn't find a reference in the sheet. It should be taught in some workshop. In that case do you think I should change the user stories and ask to use `findIndex` instead?

zairahira: At this point, `indexOf` has been introduced according to the sheet.

Dario-DC: `indexOf` does not work here because we need to search for the object key inside an array of objects.

moT01 (approved):

PR #492, opened 2024-07-24
feat: add lunch picker lab (JS only)

### LINKED ISSUE #428: chore: add lab for arrays to frontend cert (JS only)
There will be a workshop and lab that will teach campers about arrays and basic array methods.
We need a lab project for them to do right after the workshop.

Campers will know the following so far:

- let
- const
- assignment
- reassignment
- basic string usage
- console.log
- camelCase
- template literals
- str.length property
- str.indexOf method
- if/else
- Number()
- Math.floor
- Math.random
- prompt()
- ===
- if/else if/else
- comparison operators
- array basics
- push
- pop
- shift
- unshift

## Acceptance criteria

zairahira:
Checklist:

Closes #428

### DIFF
--- a/frontend-cert/js-projects/js-lunch-picker/script.js
+++ b/frontend-cert/js-projects/js-lunch-picker/script.js
@@ -1,19 +1,34 @@
 
-// task 1: declare these arrays:
+// task 1: create a kids menu with one item
+let kidsMeal = ["Chicken Nuggets"];
 
-let kidsMeal = ["Chicken Nuggets", "Pizza", "Burgers", "Tacos", "Fries", "Prawns"];
-let adultsMeal = ["Salad", "Wraps", "Hummus", "Rice"];
+// task 2: create an adults menu with one item
+let adultsMeal = ["Salad"];
 
-// task 2: remove "prawns" from kidsMeal using pop()
+// task 3: add three more items to kids menu
+kidsMeal.push("Pizza", "Burgers", "Tacos");
 
+// task 4: add three more items to adults menu
+adultsMeal.push("Wraps", "Hummus", "Rice");
+
+// task: print both menus
+console.log("Kids Menu:", kidsMeal);
+console.log("Adults Menu:", adultsMeal);
+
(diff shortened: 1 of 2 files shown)


Dario-DC (commented):
I think this could work. But what if we ask to change the `meals` array in some way instead?
Something like pushing another couple of elements to the array (let's say they're the daily specials, for example) and/or removing the chosen element. Not sure if they know how to remove elements by index or value, yet.

zairahira:
According to the sheet, the array methods like removing and adding elements come after this lab.
However, at this point campers would know how to access and update items. Should I make changes according to that?

Dario-DC:
There's a mismatch between the issue related to this PR and the sheet apparently. Unless the array methods are going to be introduced in the previous workshop (which I think has not been created yet) and then taught in the following videos.

Is that possible? @zairahira @jdwilkin4

jdwilkin4:
I went ahead and updated it on the sheet.
So push, pop, shift and unshift will be taught before this in the lecture block

jdwilkin4:
I just added the workshop PR that will go before this lab

https://github.com/freeCodeCamp/CurriculumExpansion/pull/495

zairahira:
@Dario-DC I have added a few more methods like `push()` and `pop()`. Let me know what you think of this one.

moT01:
Just throwing out ideas - you could start with an empty array, or maybe just one item in it, and have them push several more instead of just one.

"You should create a `adultsMenu` array with one item, "Salad", in it.
"You should use "push" to add three more food items to the `adultsMenu` array"
"You should create a `kidsMenu` array with one item, "Pizza", in it.
"You should use "push" to add two more food items to the `kidsMenu` array"

zairahira:
@moT01 I have added a few more stories as suggested. Let me know what you think.

Dario-DC (commented):
Sorry Zaira, I missed your comment. I think it works well :+1:

moT01 (approved):

PR #493, opened 2024-07-24
feat: adding sentence analyzer loops project

### LINKED ISSUE #425: chore: add workshop to teach loops in frontend cert(JS only)
The frontend cert will have a module just for loops.
There will be a lecture block going into more detail about how to work with loops with examples.
But we need a short(like 40 steps or less) workshop, for them to learn about loops interactively.

This workshop does not need to cover all the different types of loop because the lecture before this is covering a lot and providing them an introduction. 
So we could just focus on a few like the `while` and `for...of` loop for example.

- let
- const
- assignment
- reassignment
- basic string usage
- console.log
- camelCase
- template literals
- str.length property
- str.indexOf method
- Number()
- Math.floor
- Math.random
- prompt()
- ===
- if/else if/else
- comparison operators(< >=)
- logical operators (|| &&)
- array basics
- push
- pop
- shift
- unshift
- basic object usage
- for loop
- for...of
- for...in
- do...while
- while

## Acceptance criteria

Ksound22:
Checklist:

Closes #425

### DIFF
--- a/frontend-cert/js-projects/sentence-analyzer/script.js
+++ b/frontend-cert/js-projects/sentence-analyzer/script.js
@@ -1,0 +1,69 @@
+let sentence = prompt('Enter a sentence to analyze:');
+
+// Initialize stat variables to 0
+let wordCount = 0;
+let vowelCount = 0;
+let consonantCount = 0;
+let punctuationCount = 0;
+
+// Define varaibles for vowels, consonants, and possible punctuation marks from the sentence
+const vowels = 'aeiou';
+const consonants = 'bcdfghjklmnpqrstvwxyz';
+const punctuations = '.,!?;:-()[]{}"\'–';
+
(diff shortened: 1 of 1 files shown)


frontend-cert/js-projects/sentence-analyzer/script.js
```diff
@@ -0,0 +1,69 @@
...
+
+  if (char === ' ' || i === sentence.length - 1) {
+    wordCount++;
+  }
+}
+*/
+
+// Results
+// We could get fancy by making the sentence introducing the results bigger
+// console.log('%cSentence Analysis Results', 'font-size: 20px');
```
jdwilkin4, line 63: we could leave this out IMO

jdwilkin4 (approved):
tested it and looks good 👍

PR #502, opened 2024-07-31
feat: add factorial lab

### LINKED ISSUE #430: chore: add lab for loops in frontend cert(JS only)
The previous workshop and lecture videos will teach campers about loops.

So we need a lab for them to do immediately after that.

- let
- const
- assignment
- reassignment
- basic string usage
- console.log
- camelCase
- template literals
- str.length property
- str.indexOf method
- Number()
- Math.floor
- Math.random
- prompt()
- ===
- if/else if/else
- comparison operators(< >=)
- logical operators (|| &&)
- % operator
- array basics
- push
- pop
- shift
- unshift
- basic object usage
- for loop
- for of loop
- do/while loop
- while loop
- forEach loop

## Acceptance criteria

zairahira:
Checklist:

Closes #430

### DIFF
--- a/frontend-cert/js-projects/lab-factorial/user-stories.md
+++ b/frontend-cert/js-projects/lab-factorial/user-stories.md
@@ -1,0 +1,8 @@
+1. You should ask the user for a number using a prompt window and store the number in a variable called `inputNumber`.
+   
+2. Declare a variable `result` and assign it the value of 1.
+   
+3. Using a for loop, you should loop through all numbers from `1` to `inputNumber` (inclusive) and for each number, multiply the `result` variable by the current number.
+   
+4. You should log the result of the factorial of `inputNumber` to the console. The output should be in the format `The factorial of [inputNumber] is [result]`.
+   
--- a/frontend-cert/js-projects/lab-factorial/user-stories.md
+++ b/frontend-cert/js-projects/lab-factorial/user-stories.md
@@ -1,8 +1,8 @@
 1. You should ask the user for a number using a prompt window and store the number in a variable called `inputNumber`.
    
-2. Declare a variable `result` and assign it the value of 1.
+2. Declare a variable `result` and assign it the value of `1`.
    
-3. Using a for loop, you should loop through all numbers from `1` to `inputNumber` (inclusive) and for each number, multiply the `result` variable by the current number.
+3. Using a for loop, you should loop through all numbers from `1` to `inputNumber` (inclusive) and for each number, multiply the `result` variable by the current number and assign the result to the `result` variable.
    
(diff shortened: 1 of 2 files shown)


frontend-cert/js-projects/lab-factorial/user-stories.md
```diff
@@ -0,0 +1,8 @@
+1. You should ask the user for a number using a prompt window and store the number in a variable called `inputNumber`.
+   
+2. Declare a variable `result` and assign it the value of 1.
```
Dario-DC, line 3:
```suggestion
2. Declare a variable `result` and assign it the value of `1`.
```

frontend-cert/js-projects/lab-factorial/user-stories.md
```diff
@@ -0,0 +1,8 @@
...
+   
+3. Using a for loop, you should loop through all numbers from `1` to `inputNumber` (inclusive) and for each number, multiply the `result` variable by the current number.
```
Dario-DC, line 5:
```suggestion
3. Using a for loop, you should loop through all numbers from `1` to `inputNumber` (inclusive) and for each number, multiply the `result` variable by the current number and assign the result to the `result` variable.
```
It may not be necessary, though.

frontend-cert/js-projects/lab-factorial/user-stories.md
```diff
@@ -0,0 +1,8 @@
...
+   
+4. You should log the result of the factorial of `inputNumber` to the console. The output should be in the format `The factorial of [inputNumber] is [result]`.
```
Dario-DC, line 7:
```suggestion
4. You should log the result of the factorial of `inputNumber` to the console. The output should be in the format `The factorial of <inputNumber> is <result>`.
```
I'm not sure if this way is better, but when we ask to follow a certain format, I'd like to keep the same placeholder format everywhere, for consistency.

Dario-DC (commented):
Left a few comments. But it's not essential to implement them. The lab is looking good :+1:

frontend-cert/js-projects/lab-factorial/user-stories.md
Dario-DC: By the way, I'm okay with changing the projects in which we use `<>` into `[]`. It works either way for me.

zairahira: We can decide one format for consistency.

moT01 (approved):
LGTM 🎉

frontend-cert/js-projects/lab-factorial/user-stories.md
moT01: I think I've typically used the <>, but I think the brackets might be better. If the <> aren't inside a code block, I think they will be interpreted as HTML if we are rendering the markdown.

PR #505, opened 2024-08-02
feat: add regex lab

### LINKED ISSUE #458: chore: add lab for regex in frontend cert
We need a lab for them to do on Regular expressions.
https://www.freecodecamp.org/learn/javascript-algorithms-and-data-structures-v8/learn-regular-expressions-by-building-a-spam-filter/step-35

## Acceptance criteria

Dario-DC:
Checklist:

Closes ####

image

### DIFF
--- a/frontend-cert/js-projects/regex-sandbox/index.html
+++ b/frontend-cert/js-projects/regex-sandbox/index.html
@@ -1,0 +1,45 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <link rel="stylesheet" href="styles.css">
+    <title>Regex Sandbox</title>
+
+</head>
+
+<body>
+    <h1>Regex Sandbox</h1>
+    <main>
+        <div id="regex-container">
+            <label for="pattern">Regex Pattern:
+                <div id="pattern-container">/<input type="text" id="pattern" name="pattern"
+                        placeholder="Enter your regex pattern">/</div>
+            </label>
+            <div id="flags-container">
+                <p>Flags: </p>
+                <label for="i">
+                    <input type="checkbox" name="flags" id="i"> i
+                </label>
+                <label for="g">
+                    <input type="checkbox" name="flags" id="g"> g
+                </label>
+            </div>
+        </div>
+        <div id="test-container">
+            <p>Test String:</p>
+            <div id="test-string" placeholder="Enter your test string" contenteditable="true"></div>
+        </div>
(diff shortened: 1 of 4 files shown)


moT01:
Is this one ready?

zairahira:
@Dario-DC 
I am not sure if `matchAll` has been taught yet. I see `match` in the sheet.
I would suggest to include `matchAll` in the lecture videos and keep this project as it is quite appropriate for this lab.

The same goes for `replaceAll`, I couldn't find it in the sheet.

Dario-DC:
> @Dario-DC I am not sure if `matchAll` has been taught yet. I see `match` in the sheet. I would suggest to include `matchAll` in the lecture videos and keep this project as it is quite appropriate for this lab.
> 
> The same goes for `replaceAll`, I couldn't find it in the sheet.

They've been added to the lecture scripts.

moT01 (commented):
This is an interesting looking project, looks pretty fun - but I'm wondering if it's aligned with what we want it to teach. I don't really see a whole lot of regex that users have to write here.

Wondering if we should try to come up with something else?

Dario-DC:
I had the same doubts. I'll try to come up with something else that require campers to write regular expressions.

Do you think this could be used in any other sections of the curriculum? @moT01

Dario-DC:
Some time ago Zaira suggested to code a tool that scans a markdown file and extracts every link. This could be an idea. The markdown could be provided as a string in the seed code. What do you think? @moT01

moT01:
> Do you think this could be used in any other sections of the curriculum?

Probably. There's a number of things it covers.

> a tool that scans a markdown file and extracts every link

That's not bad, but I think you might only need to write one regex for that. Which works. But maybe something where they need to write several. I asked GPT for a list of ideas, here's a few that sound pretty decent:

Log File Analyzer: Create a script that parses a log file and extracts information like IP addresses, dates, or error messages using regex.

Simple Markdown Parser: Build a basic parser that converts simple markdown (like **bold**, *italic*, and # heading) into HTML using regex.

Password Strength Indicator: Create a password strength meter that evaluates a password based on criteria like the presence of uppercase letters, numbers, special characters, etc., using regex.

Form Validator: Create a form with fields like email, phone number, and zip code. Use regex to validate the input before submission.

I like the first two.

Dario-DC:
>That's not bad, but I think you might only need to write one regex for that.

That's true. I like the markdown parser. I'll update this PR as soon as I have something that works.

The form validator would be for another lab.

moT01:
So we will need to update the order of the blocks on there as well as in the repo. I will see if I can come up with a script to update the repo.

moT01 (approved):

PR #506, opened 2024-08-02
feat: add-library-manager-hoc-project

### LINKED ISSUE #443: chore: add workshop for higher order functions (JS only)
The frontend cert will have a module just for higher order functions.
There will be a lecture block going into more detail about how to work with callbacks and higher order functions with examples.
But we need a short(like 40 steps or less) workshop, for them to learn about higher order functions interactively.

This workshop does not need to cover all types of popular higher order functions (map, filter, reduce)
I would suggest just picking one or two of them.

- let
- const
- assignment
- reassignment
- basic string usage
- console.log
- camelCase
- template literals
- str.length property
- str.indexOf method
- Number()
- Math.floor
- Math.random
- prompt()
- ===
- if/else if/else
- comparison operators(< >=)
- logical operators (|| &&)
- array basics
- push
- pop
- shift
- unshift
- basic object usage
- for loop
- for of loop
- do/while loop
- while loop
- forEach loop
- regular functions
- arrow functions

## Acceptance criteria

Ksound22:
Checklist:

Closes #443

### DIFF
--- a/frontend-cert/js-projects/library-manager/script.js
+++ b/frontend-cert/js-projects/library-manager/script.js
@@ -1,10 +1,10 @@
 // Initialize the library with some books – we can make campers fill in the first two, then supply the rest ourselves.
 const library = [
   {
-    title: 'Battle Cry of Freedom: The Civil War Era',
-    author: 'James M. McPherson',
-    about: 'A book about the American civil war',
-    pages: 904,
+    title: 'Your Next Five Moves: Master the Art of Business Strategy',
+    author: 'Patrick Bet-David and Greg Dinkin',
+    about: 'A book on how to plan ahead',
+    pages: 320,
   },
   {
     title: 'Atomic Habits',
@@ -14,11 +14,12 @@ const library = [
     pages: 320,
   },
   {
-    title: '1776',
-    author: 'David McCullough',
+    title:
+      'Choose Your Enemies Wisely: Business Planning for the Audacious Few',
+    author: 'Patrick Bet-David',
     about:
-      'A book about those who marched with George Washington during the Declaration of Independence',
-    pages: 386,
+      "A book that emphasizes the importance of identifying and understanding one's adversaries to succeed in the business world",
+    pages: 304,
   },
   {
     title: 'The Embedded Entrepreneur',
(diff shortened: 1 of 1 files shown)


frontend-cert/js-projects/library-manager/script.js
```diff
@@ -0,0 +1,97 @@
...
+    author: 'Walter Rodney',
+    about:
+      'The history of African and Aisan immigration into Guyana and the impact of colonization on Guyana',
+    pages: 312,
+  },
+];
+
+// Add book with push()
+function addBook(title, author, pages) {
+  const newBook = { title, author, pages };
```
jdwilkin4, line 60: we will need to add a comment here about this syntax.
it might show up in one of the lecture videos before this, but we will want to re introduce here too for extra reinforcement

We could probably just have a sentence explaining the object shorthand notation and a short example and that should suffice

frontend-cert/js-projects/library-manager/script.js
```diff
@@ -0,0 +1,97 @@
...
+  library.push(newBook);
+}
+
+// Use map to send the books into an array
+const bookSummaries = library.map((book) => {
+  return `${book.title}, ${
+    book.about.charAt(0).toLowerCase() + book.about.slice(1)
```
jdwilkin4, line 67: it looks like we have a lecture video in the strings module called "What is ASCII and how does it work with charCodeAt() and fromCharCode()?"

but right now, we are not teaching the charAt method.
So you will need to add a note to teach it with an example

frontend-cert/js-projects/library-manager/script.js
```diff
@@ -0,0 +1,97 @@
...
+    about: 'A book about the analysis of precolonial African states',
+    pages: 240,
+  },
+  {
+    title: 'How Europe Underdeveloped Africa',
+    author: 'Walter Rodney',
+    about:
+      'A book detailing how European countries divided Africa between themselves and exploited it',
+    pages: 312,
+  },
```
jdwilkin4, line 48: I vote we remove this book.

It looks like a good book but unfortunately people are sensitive nowadays.
So it is best to be on the safe side here

jdwilkin4 (commented):
I think it looks good.
Left a couple of small comments 👍

frontend-cert/js-projects/library-manager/script.js
```diff
@@ -0,0 +1,97 @@
...
+    book.about.charAt(0).toLowerCase() + book.about.slice(1)
+  }`;
+});
+
+// Display books with forEach()
+function displayBooks() {
+  console.log('Library Books:');
+  library.forEach((book) => {
+    console.log(
+      `${book.title} by ${book.author}, ${book.about} (${book.pages} pages)`
```
jdwilkin4, line 76: also, let's add a newline here to make it easier to read in the console

```suggestion
      `${book.title} by ${book.author}, ${book.about} (${book.pages} pages)\n`
```

frontend-cert/js-projects/library-manager/script.js
Ksound22: I see. I will remove all the historical books there.

jdwilkin4 (approved):

PR #541, opened 2024-08-28
feat: add accessibility lab prototype

moT01:
It doesn't cover a whole lot of accessibility things, but it has a few - alt text, labels, semantic HTML.

Checklist:

Closes #XXXXX


zairahira:
The prototype is looking great overall.

Here are some suggestions from my end:
- There should be user stories for `required` and `aria-required="true"` properties.
- Although not necessary, but to elaborate the lab a bit, it would be good to add something about keyboard accessibility: either `tabindex` or `accesskey`.

PR #566, opened 2024-09-05
feat: add event flyer lab

### LINKED ISSUE #420: chore: add Absolute and Relative CSS Units Workshop
Right now, we teach absolute and relative units throughout the entire RWD cert but never clearly explain the differences between the to or when to use which unit is which situation. A lot of times we tell them, this is a px add it here. Or this is a rem, add it here. And we are seeing the results in their certification projects where they are just using whatever unit and not really understanding what is does or why they should use it in this case.

So in this new cert there will be a short module just for absolute and relate units.
We need a short workshop project teaching this.

You don't need to include all of the absolute and relative units in this workshop.
Just include a few so they can practice what they just learned in the preceding lecture block which goes into more detail about units. 

For context, this project will go before the colored markers ones
https://www.freecodecamp.org/learn/2022/responsive-web-design/learn-css-colors-by-building-a-set-of-colored-markers/step-94

So they will know everything taught in the Cafe menu project, and cat photo app projects.
In this workshop, only used named colors because they haven't leaned how to work with other color models like rgb, hsl, and hex yet.

moT01:
Checklist:

Closes


jdwilkin4:
So the issue you linked is for a workshop.

So you can unlink that, because we will still need to create a workshop if this is supposed to be the lab that goes after it

jdwilkin4:
Looking at the sheet, we could just have a lab in this module and no workshop.
Same goes for the pseudo classes and  elements module that follows this.

Typically, we have the lecture ==> workshop ==> lab  ==> quiz order.
But there are times where we deviate from when necessary.

So this might be one of those times.

I feel like the lecture videos do a good job of explaining the concepts and we could use the labs as there way to practice what they just learned. So we might not need the workshops too.

what do you think about this change?

frontend-cert/css-projects/event-flyer/index.html
```diff
@@ -0,0 +1,59 @@
...
+  <link rel="stylesheet" href="styles.css">
+</head>
+
+<body>
+
+  <header>
+    <img src="rock.jpg">
+    <h1>Pet Rock Festival</h1>
+    <p>Join us for the first annual Pet Rock Festival!</p>
+    <p><b>When:</b> October 3rd <b>Where:</b> Boulder Park</p>
```
jdwilkin4, line 17: I would use the `time` element for the date here

jdwilkin4 (approved):
Played around with it locally and it looks good on my end. 👍 
Left one comment about the workshops and the html file

moT01:
> just have a lab in this module and no workshop.

I'm not against it. I think the workshops are nice - I would lean towards having one if we can. But there's a lot of work to do, so maybe skip it? We can also potentially add one later - maybe if we get time later this fall.

jdwilkin4:
> We can also potentially add one later - maybe if we get time later this fall.

We can take that approach.

So highest priority will be placed on the remaining labs and existing workshops, and then if time permits we will add in the two css workshops. Otherwise, we will ship the first version without it 👍

PR #585, opened 2024-09-12
feat: add second OOP lab

### LINKED ISSUE #470: chore: add second lab for OOP in frontend cert
We need a second lab for them to do help them review working with OOP.
https://www.freecodecamp.org/learn/javascript-algorithms-and-data-structures-v8/learn-intermediate-oop-by-building-a-platformer-game/step-117

## Acceptance criteria

- [ ] add your completed project in that new folder

ihechikara:
Checklist:

Closes #470

### DIFF
--- a/frontend-cert/js-projects/second-oop/user-stories.md
+++ b/frontend-cert/js-projects/second-oop/user-stories.md
@@ -1,38 +1,13 @@
-1. You should define a class named `BankAccount` with a `constructor` that has a `balance` parameter (default value of 0). The constructor should also initialize an empty `transactions` array to store transaction records.
+1. You should define a class named `BankAccount` with a `constructor` that has a `balance` parameter. The default parameter value should be set to 0. The constructor should also initialize an empty `transactions` array to store transaction records.
 
-2. You should define a method named `deposit` with an `amount` parameter. This method should:
-    - Check if the `amount` is greater than 0.
-    - If the amount is valid, push an object with the properties `type: 'deposit'` and the `amount` into the `transactions ` array.
-    - Call the `updateBalance` method to update the account's balance.
-    - Log a message to the console indicating the deposit was successful and display the new balance.
-    - If the `amount` is invalid, log an error message indicating that the deposit amount must be greater than zero.
+2. You should define a method named `updateBalance`.
 
-3. You should define a method named `withdraw` with an `amount` parameter. This method should:
-    - Check if the `amount` is greater than 0 and less than or equal to the current `balance`.
-    - If valid, push an object with the properties `type: 'withdraw'` and the `amount` into the `transactions` array.
-    - Call the `updateBalance` method to update the account's balance.
-    - Log a message to the console indicating the withdrawal was successful and display the new balance.
-    - If the `amount` is invalid or exceeds the balance, log an error message indicating insufficient balance or invalid amount.
+3. You should define a method named `deposit` with an `amount` parameter. This method should update `balance` accordingly.
 
-4. You should define a method named `checkBalance` that logs the current account balance to the console.
+4. You should define a method named `withdraw` with an `amount` parameter. This method should update `balance` accordingly.
 
-5. You should define a method named `updateBalance`. This method should:
-    - Use the `reduce` function to sum the transactions in the `transactions` array.
-    - Update the `balance` by adding up all the deposit amounts and subtracting all the withdrawal amounts.
+5. You should define a method named `checkBalance` that returns the current balance.
 
-6. You should define a method named listAllDeposits. This method should:
-    - Use the `filter` function to create a new array containing only the transactions with `type: 'deposit'`.
-    - Use the `map` function to extract the `amount` from each deposit transaction.
-    - Log the list of deposit amounts to the console and return this list.
+6. You should define a method named `listAllDeposits` that returns all deposits.
 
-7. You should define a method named `listAllWithdrawals`. This method should:
-    - Use the `filter` function to create a new array containing only the transactions with `type: 'withdraw'`.
-    - Use the `map` function to extract the `amount` from each withdrawal transaction.
-    - Log the list of withdrawal amounts to the console and return this list.
-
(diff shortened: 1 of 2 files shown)


fhsinchy (approved):

frontend-cert/js-projects/second-oop/user-stories.md
```diff
@@ -0,0 +1,38 @@
+1. You should define a class named `BankAccount` with a `constructor` that has a `balance` parameter (default value of 0). The constructor should also initialize an empty `transactions` array to store transaction records.
```
jdwilkin4, line 1:
```suggestion
1. You should define a class named `BankAccount` with a `constructor` that has a `balance` parameter. The default parameter value should be set to `0`. The constructor should also initialize an empty `transactions` array to store transaction records.
```

frontend-cert/js-projects/second-oop/user-stories.md
```diff
@@ -0,0 +1,38 @@
...
+
+2. You should define a method named `deposit` with an `amount` parameter. This method should:
+    - Check if the `amount` is greater than 0.
+    - If the amount is valid, push an object with the properties `type: 'deposit'` and the `amount` into the `transactions ` array.
+    - Call the `updateBalance` method to update the account's balance.
+    - Log a message to the console indicating the deposit was successful and display the new balance.
+    - If the `amount` is invalid, log an error message indicating that the deposit amount must be greater than zero.
```
jdwilkin4, line 8: this is too step by step and doesn't allow the camper to think through what needs to be done.
this should be more generalized so the camper is forced to think about how to solve the problem

jdwilkin4: Also, it will be much easier from a testing perspective to return the string instead of logging it to the console.

frontend-cert/js-projects/second-oop/user-stories.md
```diff
@@ -0,0 +1,38 @@
...
+
+3. You should define a method named `withdraw` with an `amount` parameter. This method should:
+    - Check if the `amount` is greater than 0 and less than or equal to the current `balance`.
+    - If valid, push an object with the properties `type: 'withdraw'` and the `amount` into the `transactions` array.
+    - Call the `updateBalance` method to update the account's balance.
+    - Log a message to the console indicating the withdrawal was successful and display the new balance.
+    - If the `amount` is invalid or exceeds the balance, log an error message indicating insufficient balance or invalid amount.
```
jdwilkin4, line 15: same issue here.
This should be more generalized to have camper think more about what needs to be done and not so step by step.

frontend-cert/js-projects/second-oop/user-stories.md
```diff
@@ -0,0 +1,38 @@
...
+
+4. You should define a method named `checkBalance` that logs the current account balance to the console.
```
jdwilkin4, line 17: we should just have them return the string instead of logging it.
it will be much easier to test that way

frontend-cert/js-projects/second-oop/user-stories.md
```diff
@@ -0,0 +1,38 @@
...
+
+5. You should define a method named `updateBalance`. This method should:
+    - Use the `reduce` function to sum the transactions in the `transactions` array.
+    - Update the `balance` by adding up all the deposit amounts and subtracting all the withdrawal amounts.
```
jdwilkin4, line 21: this one too should be more generalized.
we should allow for multiple valid answers but right now we are only asking them to user reduce but there are multiple ways to solve this

frontend-cert/js-projects/second-oop/user-stories.md
```diff
@@ -0,0 +1,38 @@
...
+
+6. You should define a method named listAllDeposits. This method should:
+    - Use the `filter` function to create a new array containing only the transactions with `type: 'deposit'`.
+    - Use the `map` function to extract the `amount` from each deposit transaction.
+    - Log the list of deposit amounts to the console and return this list.
```
jdwilkin4, line 26: we should be more general here to and say something like "Return an array containing the amounts from each deposit transaction."

also, we should remove the log statement here.

frontend-cert/js-projects/second-oop/user-stories.md
```diff
@@ -0,0 +1,38 @@
...
+
+7. You should define a method named `listAllWithdrawals`. This method should:
+    - Use the `filter` function to create a new array containing only the transactions with `type: 'withdraw'`.
+    - Use the `map` function to extract the `amount` from each withdrawal transaction.
+    - Log the list of withdrawal amounts to the console and return this list.
```
jdwilkin4, line 31: this could be updated to something like "Return an array containing the amounts from each withdrawal transaction."

also, we should remove the log statement here.

frontend-cert/js-projects/second-oop/user-stories.md
```diff
@@ -0,0 +1,38 @@
...
+
+8. You should create an instance of the BankAccount class named myAccount. Then, you should:
+    - Use the `deposit` method to deposit into the account.
+    - Use the `withdraw` method to withdraw from the account.
+    - Use the `checkBalance` method to log the current balance to the console.
+    - Use the `listAllDeposits` method to list all deposits made.
+    - Use the `listAllWithdrawals` method to list all withdrawals made.
```
jdwilkin4, line 38: IMO, I don't think we really need this last part.

jdwilkin4 (commented):
Hey @ihechikara ! 

I think the project itself works well but the user stories overall are to strict and step by step.

Remember the goal of the labs is for campers to build things on their own and think through the problems. That is the only way you grow and learn as a developer.

If everything is prescribed to them in a step by step fashion, then they don't really need to think through what to do because they are just going through the motions of a tutorial.

I left some comments on where we should generalize things to make this more open ended so it matches what is currently merged into main.

ihechikara:
Thanks for the feedback, @jdwilkin4. I've updated the user stories. Let me know what you think.

jdwilkin4 (approved):
Yes, this is much more opened ended 👍

PR #616, opened 2024-09-23
chore: recursion lab project

### LINKED ISSUE #464: chore: add recursion lab to frontend cert
We need a lab for them to do help them review working with recursion.
https://www.freecodecamp.org/learn/javascript-algorithms-and-data-structures-v8/learn-recursion-by-building-a-decimal-to-binary-converter/step-109

## Acceptance criteria

larymak:
Checklist:

Closes #464


frontend-cert/js-projects/lab-recursion/tempCodeRunnerFile.js
```diff
@@ -0,0 +1 @@
+n
```
jdwilkin4, line 1: this code runner file can be removed

jdwilkin4 (commented):
We already have a factorial lab merged into main here

https://github.com/freeCodeCamp/freeCodeCamp/pull/55889

we also have an nth fibonacci lab here 
https://github.com/freeCodeCamp/CurriculumExpansion/pull/490

Any of these options are open

- tower of hanoi
- permutations of a string using recursion

or any other options you can think of that aren't on the labs sheet 👍

jdwilkin4 (commented):
Your function should return the permutations. 
Right now, the only return value is if the str is empty. So it is only returning the prefix in that case. 

but I should be able to call the function like this: `console.log(permuteString('ABC'));` and get back all of the possible permutations like this: `[ 'ABC', 'ACB', 'BAC', 'BCA', 'CBA', 'CAB' ]`

or call it with str `XY` and have the function return `[“XY”, “YX”]`

frontend-cert/js-projects/lab-recursion/user-stories.md
```diff
@@ -0,0 +1,7 @@
+1. You should create a function named `permuteString`.
+1. The `permuteString` function should take three parameters, `str` which is the string to be permuted, an optional parameter `prefix` which defaults to an empty string and an empty array named `result`.
+1. You should check if the length of the `str` is 0, if it is, you should push the `prefix` value to the `result` array and return the `result` array.
+1. You should create a for loop to iterate over each character in the `str` aganist the `str` length and increment the index by 1.
+1. You should create a new string named `rem` which is the concatenation of the substring of `str` from the beginning up to, but not including, the current index, and the substring of `str` from the character right after the current index to the end.
+1. You should recursively call the `permuteString` function with the `rem` string as the first argument, the concatenation of the `prefix` and the character at the current index as the second argument, and the `result` array as the third argument.
```
jdwilkin4, line 6: We should take a more generalized approach here. 
As it stands now, I am not having to think to much on what to do here because it tells me what to increment, what variables to create, what methods to use, etc.

Having things be more generalized, forces campers to think more about what to do and gives them more freedom to implement their own solution. This makes for a much better learning experience and strengthens their problem solving skills

maybe something like this:

```md
Check if the length of str is 0. If it is, push prefix to result and return result.
Implement a loop to iterate over each character in str.
Recursively call the permuteString function with the appropriate arguments to generate permutations.
```

jdwilkin4 (commented):
Left some comments on the user stories

jdwilkin4 (approved):
I think there might be room for tweaking the user stories, but we will know more when it is added to the main repo with tests.

Let's get this in

PR #625, opened 2024-09-25
feat: replace student grade tracker with recipe tracker

### LINKED ISSUE #618: chore: update objects workshop prototype
We have this workshop on objects here
https://github.com/freeCodeCamp/CurriculumExpansion/pull/457/files

This prototype was created before we moved the function module.
So we will need to update the prototype, to include functions so it will be easier to test. 
Also, this workshop is pretty similar to the gradebook app lab we already have merged into main.
So let's update the project to use a different subject matter.

Ksound22:
Checklist:

Closes #618


jdwilkin4 (commented):
We still need to address the second part of the issue which was to update the workshop to be on a different subject matter then students and grades.

We already have a lab that is pretty similar to this where campers are building really similar functionality.

One of the idea I thought of was to have them build a recipe tracker of sorts.

there could be an array of objects like this

```js
const recipes = [
  {
    name: "Spaghetti Carbonara",
    ingredients: [
      "spaghetti",
      "eggs",
      "Parmesan cheese",
      "pancetta",
      "black pepper"
    ],
    cookingTime: 20,
    ratings: [4, 5, 4],
    averageRating: null
  },
...
```

and then there could be a function to calculate the averageRating

then there could be console.logs for ratings results, or any other information we would like to log to the console

other ideas could be a travel journal or fitness tracker.

it is up to you what subject matter you want to choose as long as it is not students and grades

Ksound22:
> We still need to address the second part of the issue which was to update the workshop to be on a different subject matter then students and grades.
> 
> We already have a lab that is pretty similar to this where campers are building really similar functionality.
> 
> One of the idea I thought of was to have them build a recipe tracker of sorts.
> 
> there could be an array of objects like this
> 
> ```js
> const recipes = [
>   {
>     name: "Spaghetti Carbonara",
>     ingredients: [
>       "spaghetti",
>       "eggs",
>       "Parmesan cheese",
>       "pancetta",
>       "black pepper"
>     ],
>     cookingTime: 20,
>     ratings: [4, 5, 4],
>     averageRating: null
>   },
> ...
> ```
> 
> and then there could be a function to calculate the averageRating
> 
> then there could be console.logs for ratings results, or any other information we would like to log to the console
> 
> other ideas could be a travel journal or fitness tracker.
> 
> it is up to you what subject matter you want to choose as long as it is not students and grades

I didn't notice we had to change everything. Should be fine now.

jdwilkin4 (commented):
Left a couple of comments but otherwise looks good

jdwilkin4 (approved):
we can add this in 👍

PR #630, opened 2024-10-07
feat: functional programming lab

### LINKED ISSUE #468: chore: add lab for functional programming in frontend cert
We need a lab for them to do help them review working with functional programming.

## Acceptance criteria

- [ ] add your completed project in that new folder

Dario-DC:
Checklist:

Closes #468 

image
image

### DIFF
--- a/frontend-cert/js-projects/functional-programming-lab/index.html
+++ b/frontend-cert/js-projects/functional-programming-lab/index.html
@@ -1,0 +1,25 @@
+<!DOCTYPE html>
+<html lang="en">
+
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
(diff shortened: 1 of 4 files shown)


jdwilkin4 (commented):
I was able to play around with it locally, and to me I don't think the implementation really shows the swaps like other visualizers do.
Screenshot 2024-10-18 at 5 44 08 AM


I feel like we should have something closer that shows when a swap is made
https://www.cs.usfca.edu/~galles/visualization/ComparisonSort.html

Dario-DC:
I initially did something closer to what you would have expected, but then I asked myself: how am I supposed to test that?
@jdwilkin4

jdwilkin4 (approved):
That's fair. 

I tried to play around with it to see if adding text when a swap occurs would work but it ended up just looking crowded in the UI.
So we can leave it as is. 👍

PR #662, opened 2024-11-06
feat: add Map and Set lab prototype

### LINKED ISSUE #654: Creat prototype for Maps and Sets lab
We need to create a prototype for a short lab for campers to practice working with Maps and and Sets.

You can review the previous lecture block on the sheet to see what concepts were taught before this.

zairahira:
Checklist:

Closes #654

### DIFF
--- a/frontend-cert/js-projects/voting-system/index.js
+++ b/frontend-cert/js-projects/voting-system/index.js
@@ -23,12 +23,12 @@ function vote(option, voterId) {
   }
 }
 
-// Display poll results
 function displayResults() {
-  console.log("Poll Results:");
-  for (let [option, voters] of poll.entries()) {
-    return `${option}: ${voters.size} votes`;
-  }
+    let results = "Poll Results:\n";
+    for (let [option, voters] of poll.entries()) {
+        results += `${option}: ${voters.size} votes\n`;
+    }
+    return results.trim();
 }
 
 // Example usage
--- a/frontend-cert/js-projects/voting-system/index.js
+++ b/frontend-cert/js-projects/voting-system/index.js
@@ -1,0 +1,48 @@
+// Initialize the poll with a Map
+const poll = new Map();
+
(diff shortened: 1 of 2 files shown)


frontend-cert/js-projects/voting-system/user-stories.md
```diff
@@ -0,0 +1,39 @@
...
+
+1. You should initialize the `poll` variable to a new `Map` object.
+   
+2. You should have a function `addOption` that accepts a parameter `option`.
+   
+3. In the `addOption` function:
+ 
+- If the `option` does not already exist in the poll, it should be added to the poll with an empty `Set` as its value to track voters. You should also return the message `Option "<option>" added to the poll.`
+
+- If the `option` already exists, it should display the message `Option "<option>" already exists.`.
```
jdwilkin4, line 17:
```suggestion
- If the `option` already exists, it should return the message `Option "<option>" already exists.`.
```

frontend-cert/js-projects/voting-system/index.js
```diff
@@ -0,0 +1,48 @@
...
+  }
+}
+
+// Display poll results
+function displayResults() {
+  console.log("Poll Results:");
+  for (let [option, voters] of poll.entries()) {
+    return `${option}: ${voters.size} votes`;
+  }
+}
```
jdwilkin4, line 32: Having this return here will stop the function execution after the first iteration. 

It would be better to have a result string and then append the results to that string and return it outside of the loop if that makes sense

jdwilkin4 (commented):
Left a couple of comments but otherwise looks good

frontend-cert/js-projects/voting-system/index.js
zairahira: Done!

jdwilkin4 (approved):
looks good

PR #718, opened 2025-03-18
chore(curriculum): multiplication table generator lab prototype

### LINKED ISSUE #692: Create a prototype for Loops and Sequences Lab

ihechikara:
Checklist:

Closes #692

### DIFF
--- a/fullstack-cert/python-projects/muliplication-table-gen-lab/user-stories.md
+++ b/fullstack-cert/python-projects/muliplication-table-gen-lab/user-stories.md
@@ -1,0 +1,8 @@
+1. You should define a function named `multiplication_table` that takes two parameters:
+   - `n` (an integer representing the number for which the multiplication table will be generated).
+   - `limit` (an integer representing the highest multiple).
+2. Inside the `multiplication_table` function, you should:
+   - Print the multiplication table being generated.
+   - Use a loop to iterate through numbers from `1` to `limit`.
+   - Print each multiplication result in the format `n × i = result`.
+3. You should prompt the user to enter a number using `Enter a number: `, convert it to an integer, and call the `multiplication_table` function with the user input.
--- a/fullstack-cert/python-projects/muliplication-table-gen-lab/main.py
+++ b/fullstack-cert/python-projects/muliplication-table-gen-lab/main.py
@@ -1,0 +1,7 @@
+def multiplication_table(n, limit=12):  
+  print(f"Multiplication Table for {n}:")  
+  for i in range(1, limit + 1):  
+    print(f"{n} × {i} = {n * i}")  
+
+num = int(input("Enter a number: "))  
+multiplication_table(num)  


fullstack-cert/python-projects/muliplication-table-gen-lab/user-stories.md
```diff
@@ -0,0 +1,8 @@
+1. You should define a function named `multiplication_table` that takes two parameters:
+   - `n` (an integer representing the number for which the multiplication table will be generated).
+   - `limit` (an integer representing the highest multiple).
+2. Inside the `multiplication_table` function, you should:
+   - Use a loop to iterate through numbers from `1` to `limit`.
+   - Print each multiplication result in the format `n × i = result`.
+   - Print the multiplication table being generated.
```
zairahira, line 7:
```suggestion
   - Print the multiplication table being generated. For example, when user enters `5`, the following should be printed to the console: `Multiplication Table for 5:`
   - Use a loop to iterate through numbers from `1` to `limit`.
   - Print each multiplication result in the format `n × i = result`.
```

ilenia-magoni:
this seems a bit semplicistic, also it uses a loop but not much sequences

Dario-DC:
Hey @ihechikara I agree with Ilenia, we should come up with a different idea for this lab.

ilenia-magoni:
I am not sure we can use dictionaries at this point of the curriculum, the lectures for these come later

Dario-DC:
As Ilenia pointed out, dictionaries comes later on in the curriculum. So we cannot use them here.

Dario-DC (commented):
Hey @ihechikara we need to change some of the requirements for this lab. We should not use `input` (this holds for labs in general) because it complicates testing.

We cannot test several things, unless we force them to use specific variables names.

While
> Use a `for` loop that goes from 1 to `n` (inclusive).`

won't be a problem, things like

>   - In each iteration, create a list of numbers from 1 to the current value of `i`.
>   - Convert each number in the list to a string and print them joined by a space.

are not easy to check. This is what I propose:

```py
def number_pattern(n):
    result = '1'
    if isinstance(n, int):
        if n > 0:
            for i in range(2, n + 1):
                result += f' {i}'
            return result
        return 'Argument must be an integer greater than 0.'
    return 'Argument must be an integer value.'
        
print(number_pattern(13))
```

```md
1. You should define a function named `number_pattern` that takes a single parameter `n` (representing a positive integer).
1. `number_pattern` should use a `for` loop.
1. `number_pattern(n)` should return a string with all the integers starting from 1 up to `n` (included) separated by a space. For example, `number_pattern(4)` should return the string `1 2 3 4`.
1. If the argument passed to the function is not an integer value, the function should return `Argument must be an integer value.`.
1. If the argument passed to the function is less than 1, the function should return `Argument must be an integer greater than 0.`.
```

majestic-owl448:
Dario proposal seems a good fit, I would go with that one

PR #747, opened 2025-04-09
chore(curriculum): robot control lab prototype

### LINKED ISSUE #706: Create a prototype for Abstraction Lab
We need to create a prototype for a lab that teaches abstraction. Here we could use the `abc` module from the standard library to create an interface.

Campers will know how to:
- work with common data types
- handle errors
- work with classes and objects
- inheritance, encapsulation, and polymorphism

ihechikara:
Checklist:

Closes #706

### DIFF
--- a/fullstack-cert/python-projects/robot-control-lab/main.py
+++ b/fullstack-cert/python-projects/robot-control-lab/main.py
@@ -1,0 +1,56 @@
+from abc import ABC, abstractmethod
+
+class Robot(ABC):
+    @abstractmethod
+    def move_forward(self):
+        pass
+
+    @abstractmethod
+    def turn(self, direction):
+        pass
+
+    @abstractmethod
+    def perform_task(self):
+        pass
+
+class CleaningRobot(Robot):
+    def move_forward(self):
+        print("CleaningRobot moves forward 1 meter.")
+
+    def turn(self, direction):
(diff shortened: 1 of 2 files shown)


Dario-DC (commented):
Hey @ihechikara as we briefly discussed in our last meeting we should implement some functionalities apart from printing stuff.

I haven't found any interesting idea so far for this program but here's what I thought. Keeping the same structure of an abstract class with a couple of abstract methods (and optionally class methods) and a couple of concrete classes, we could implement different kind of search and/or replace (case sensitive/insensitive) functionalities in different data structures (lists, dicts, plain text).

I know it's a rough idea but maybe it's a starting point.

gikf:
> I haven't found any interesting idea so far for this program but here's what I thought. Keeping the same structure of an abstract class with a couple of abstract methods (and optionally class methods) and a couple of concrete classes, we could implement different kind of search and/or replace (case sensitive/insensitive) functionalities in different data structures (lists, dicts, plain text).

For something called _robot control lab_ it's almost a requirement to consider giving robots an actual 2D area, on which they can move around and perform their duties.

fhsinchy:
I initially came up with an idea of a file writer and parser that can write and parse the different types. But then I remembered that we can not write files on the platform. Still trying out different ideas.

PR #758, opened 2025-04-23
feat: add medical data validator (dictionary and sets workshop)

### LINKED ISSUE #693: Create a prototype for Dictionaries and Sets Workshop

Dario-DC:
Checklist:

Closes #693

The idea here is to provide a set of data in the form of a list of dictionaries (where each dictionary represents a patient) and require to validate the data checking both the general structure and that the key-value pairs of each dictionary follow the desired pattern.

I used a dictionary comprehension, which we can teach directly here without including it in the lecture IMO.

Also, I used a couple of regular expressions, nothing fancy. The standard library is introduced just before this workshop so it should be fine.

### DIFF
--- a/fullstack-cert/python-projects/medical-data-validator-workshop/main.py
+++ b/fullstack-cert/python-projects/medical-data-validator-workshop/main.py
@@ -1,0 +1,82 @@
+import re
+
+medical_records = [
+    {
+        "patient_id": "P1001",
+        "age": 34,
+        "gender": "Female",
+        "diagnosis": "Hypertension",
+        "medications": ["Lisinopril"],
+        "last_visit_id": "V2301",
+    },
+    {
+        "patient_id": "P1002",
+        "age": 47,
+        "gender": "Male",
+        "diagnosis": "Type 2 Diabetes",
+        "medications": ["Metformin", "Insulin"],
+        "last_visit_id": "V2302",
+    },
+    {
+        "patient_id": "P1003",
+        "age": 29,
+        "gender": "Female",
+        "diagnosis": "Asthma",
+        "medications": ["Albuterol"],
+        "last_visit_id": "V2303",
+    },
+    {
+        "patient_id": "P1004",
+        "age": 56,
+        "gender": "Male",
+        "diagnosis": "Chronic Back Pain",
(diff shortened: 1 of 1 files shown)


fullstack-cert/python-projects/medical-data-validator-workshop/main.py
```diff
@@ -0,0 +1,82 @@
...
+        "age": 56,
+        "gender": "Male",
+        "diagnosis": "Chronic Back Pain",
+        "medications": ["Ibuprofen", "Physical Therapy"],
+        "last_visit_id": "V2304",
+    },
+]
+
+
+def check_values(patient_id, age, gender, diagnosis, medications, last_visit_id):
```
gikf, line 39: Name of the function is not very clear, this is particularly visible in what is returned from it. Without knowing it before, I'm not sure if I'd expect list of keys with wrong values from `check_values` function. Following the usage in `validate`, one idea for better name is `find_invalid_records`.

fullstack-cert/python-projects/medical-data-validator-workshop/main.py
```diff
@@ -0,0 +1,82 @@
...
+        and all(isinstance(i, str) for i in medications),
+        "last_visit_id": isinstance(last_visit_id, str)
+        and re.fullmatch(r"V\d+", last_visit_id),
+    }
+
+    return [key for key, value in constraints.items() if not value]
+
+
+def validate(data):
+    is_list_of_dicts = isinstance(data, list) and all(isinstance(i, dict) for i in data)
```
gikf, line 56: To allow a bit more flexible types, and to limit Python's duck typing a bit less, instead of `list` and `dict`, could be used `Sequence` and `Mapping` from `collections.abc` module.

Dario-DC: I'm not sure. `Sequence` includes lists, tuples, but also ranges and strings, which should not be allowed in this case. As for `Mapping`, we are not mentioning other mappings other than dictionaries. So my preference is to keep using just `list`  (or `list` and `tuple`) and `dict`.

fhsinchy (approved):

PR #766, opened 2025-04-29
feat: add abstraction workshop

### LINKED ISSUE #705: Create a prototype for Abstraction Workshop
We need to create a prototype for a workshop that teaches abstraction and polymorphism. Here we could use the `abc` module from the standard library to create an interface.

Campers will know how to:
- work with common data types
- handle errors
- work with classes and objects
- inheritance, encapsulation, and polymorphism

larymak:
Checklist:

Closes #705


larymak:
@Dario-DC does this workshop need a `user-stories.md` file?

Dario-DC:
@larymak This seems a bit too simple. We need something that has a certain degree of complexity.

zairahira:
Hi @larymak, the workshop seems a bit basic as compared to previous labs and workshops.

fhsinchy:
I agree with @Dario-DC. Here's an idea that I came up with this morning. It's a rule based discount engine. Feel free to take a look - https://gist.github.com/fhsinchy/7d58a4b8414cf480a70176f10e6865a6

larymak:
I already updated it with a new project

Dario-DC:
@larymak the idea proposed by @fhsinchy seems to fit better. The implementation of inherited methods differs in the concrete classes. Technically, it is different in your prototype as well, Lary. But those methods print very similar strings, so I don't think it would make sense to use ABC in that case.

So I would suggest to take a look at Farhan's proposal.

fullstack-cert/python-projects/abstraction-workshop/main.py
```diff
@@ -0,0 +1,72 @@
...
+
+    def apply_discount(self, product: Product) -> float:
+        return product.price * (1 - self.percent / 100)
+
+class FixedAmountDiscount(DiscountStrategy):
+    def __init__(self, amount: float):
+        self.amount = amount
+
+    def is_applicable(self, product: Product, user_tier: str) -> bool:
+        return product.price > self.amount
```
ilenia-magoni, line 32: Here make so that any final price is fine, even 0.01
Maybe something like...

```suggestion
        return product.price*0.9 > self.amount
```

in this case the final price can't be lower than 10% of the original price

jdwilkin4:
Going to go ahead and merged this in.
If we decide not to move forward with abstraction content, then there is not harm in this project just sitting in the folder.
If we do move forward with abstraction content, then we can use this project and apply any changes as necessary.

PR #865, opened 2025-07-10
Feat/create pong prototype

### LINKED ISSUE #790: Add prototype for pong game
We are planning on using this for the revamped coding challenges section.

https://www.freecodecamp.org/learn/coding-interview-prep/take-home-projects/build-a-pong-game

The demo project works fine, but it is currently hosted on another contributor's codepen. We should host all of our own solutions on the demo projects repo.

Create a new prototype and design

Here are some ideas.

## start screen

Image

## game screen

Image

a2937:
Checklist:

Closes #790

My take on the project mostly works. The computer player's paddle goes up and down unnaturally; and I'm not quite so sure how to fix it. 

Additionally the ball sometimes gets stuck on the paddle; thus causing a glitch where the player scores.

### DIFF
--- a/coding-challenges/pong-game/index.html
+++ b/coding-challenges/pong-game/index.html
@@ -1,0 +1,29 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+  <meta charset="UTF-8">
+  <meta name="viewport" content="width=device-width, initial-scale=1.0">
+  <title>Pong</title>
+  <link rel="stylesheet" href="./styles.css"/>
+</head>
+<body>
+  <h1>Pong Game</h1>
+  <div id="game-rules" class="game-rules center">
+    <h3>Pong Challenge</h3>
+    <p>Try to score against the unbeatable computer opponent!</p>
+    <p>Move your paddle with your mouse or touch</p>
+    <button id="start-game" class="start-button" onclick="startGame()">Start Game</button>
+  </div>
+  <div id="game" class="hidden">   
+      <p id="player-score"></p>
+      <p id="computer-score"></p>        
+    <div class="board center"> 
+    <div class="vertical_dotted_line"></div>     
+      <p class="player_1 paddle"></p>
+      <p class="player_2 paddle"></p>
+    </div>
+    <button id="reset" class="reset-button">Reset Game</button>
+  </div>
+  <script src="./script.js"/> 
+</body>
+</html>
--- a/coding-challenges/pong-game/index.html
+++ b/coding-challenges/pong-game/index.html
@@ -1,29 +1,38 @@
-<!DOCTYPE html>
+<!doctype html>
 <html lang="en">
-<head>
-  <meta charset="UTF-8">
-  <meta name="viewport" content="width=device-width, initial-scale=1.0">
-  <title>Pong</title>
-  <link rel="stylesheet" href="./styles.css"/>
-</head>
-<body>
-  <h1>Pong Game</h1>
-  <div id="game-rules" class="game-rules center">
-    <h3>Pong Challenge</h3>
-    <p>Try to score against the unbeatable computer opponent!</p>
-    <p>Move your paddle with your mouse or touch</p>
-    <button id="start-game" class="start-button" onclick="startGame()">Start Game</button>
-  </div>
-  <div id="game" class="hidden">   
-      <p id="player-score"></p>
-      <p id="computer-score"></p>        
-    <div class="board center"> 
-    <div class="vertical_dotted_line"></div>     
-      <p class="player_1 paddle"></p>
-      <p class="player_2 paddle"></p>
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Pong</title>
+    <link rel="stylesheet" href="./styles.css" />
+  </head>
+  <body>
+    <h1>Pong Game</h1>
+    <div id="game-rules" class="game-rules center green-border">
+      <h3>Pong Challenge</h3>
+      <p>Try to score against the unbeatable computer opponent!</p>
+      <p>Move your paddle with your mouse or touch</p>
+      <br />
+      <button id="start-game" class="start-button" onclick="startGame()">
+        Start Game
+      </button>
     </div>
-    <button id="reset" class="reset-button">Reset Game</button>
-  </div>
-  <script src="./script.js"/> 
-</body>
+    <div id="game" class="hidden">
+      <div class="center inline">
+        <p id="player-score" class="player-score"></p>
+        <p id="computer-score" class="computer-score"></p>
+      </div>
+      <canvas
+        id="board"
+        class="board center green-border"
+        width="800"
+        height="400"
+      ></canvas>
+      <br />
+      <button id="reset" class="reset-button" onclick="resetGame()">
+        Reset Game
+      </button>
+    </div>
+    <script src="./script.js" />
+  </body>
 </html>
--- a/coding-challenges/pong-game/script.js
+++ b/coding-challenges/pong-game/script.js
@@ -1,0 +1,15 @@
+let playerScore = 0; 
+let computerScore = 0; 
+const playerOneScoreElement = document.querySelector("#player-score");
+const playerTwoScoreElement = document.querySelector("#computer-score");
+
+function startGame() {
+  const gameRules = document.querySelector(".game-rules");
+  const game = document.querySelector("#game"); 
+  gameRules?.classList.add("hidden");
+  game?.classList.remove("hidden"); 
+  // @ts-ignore
+  playerOneScoreElement.textContent = "Player:" + playerScore;
+  // @ts-ignore
+  playerTwoScoreElement.textContent = "Computer:" + computerScore;
+}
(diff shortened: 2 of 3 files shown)


jdwilkin4:
> The computer player's paddle goes up and down unnaturally; and I'm not quite so sure how to fix it.

> Additionally the ball sometimes gets stuck on the paddle; thus causing a glitch where the player scores.

I would suggest taking a look at the current solution and seeing how they approached it.

a2937:
I've not scored against the computer paddle yet. I think we're good?

a2937:
Git got weird because I was messing around with parcel the other day. My apologies

coding-challenges/pong-game/index.html
```diff
@@ -0,0 +1,38 @@
...
+    <link rel="stylesheet" href="./styles.css" />
+  </head>
+  <body>
+    <h1>Pong Game</h1>
+    <div id="game-rules" class="game-rules center green-border">
+      <h3>Pong Challenge</h3>
+      <p>Try to score against the unbeatable computer opponent!</p>
+      <p>Move your paddle with your mouse or touch</p>
+      <br />
+      <button id="start-game" class="start-button" onclick="startGame()">
```
jdwilkin4: we should remove all inline handlers since it is not a recommend practice in vanilla JS and it will match with the rest of the projects in the curriculum that use addEventListener

coding-challenges/pong-game/index.html
```diff
@@ -0,0 +1,38 @@
...
+        <p id="computer-score" class="computer-score"></p>
+      </div>
+      <canvas
+        id="board"
+        class="board center green-border"
+        width="800"
+        height="400"
+      ></canvas>
+      <br />
+      <button id="reset" class="reset-button" onclick="resetGame()">
```
jdwilkin4: same here. this inline event handler should be removed.

coding-challenges/pong-game/index.html
```diff
@@ -0,0 +1,38 @@
...
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Pong</title>
+    <link rel="stylesheet" href="./styles.css" />
+  </head>
+  <body>
+    <h1>Pong Game</h1>
+    <div id="game-rules" class="game-rules center green-border">
+      <h3>Pong Challenge</h3>
```
jdwilkin4: After looking at the initial start screen, I think it is fine to remove this completely. it looks a little repetitive because the `h1` says something really similar.

coding-challenges/pong-game/index.html
```diff
@@ -0,0 +1,38 @@
...
+      <p>Try to score against the unbeatable computer opponent!</p>
+      <p>Move your paddle with your mouse or touch</p>
```
jdwilkin4: missing punctuation but also let's remove that last part

```suggestion
      <p>Move your paddle with your mouse.</p>
```

coding-challenges/pong-game/index.html
```diff
@@ -0,0 +1,38 @@
...
+        class="board center green-border"
+        width="800"
+        height="400"
+      ></canvas>
+      <br />
+      <button id="reset" class="reset-button" onclick="resetGame()">
+        Reset Game
+      </button>
+    </div>
+    <script src="./script.js" />
```
jdwilkin4: this is not valid html. 

```suggestion
    <script src="./script.js"></script>
```

coding-challenges/pong-game/styles.css
```diff
@@ -0,0 +1,70 @@
...
+}
+
+.board {
+  border-radius: 14px;
+}
+
+.game-rules {
+  height: 15rem;
+  width: 15rem;
+  display: block;
```
jdwilkin4: this doesn't look like it is needed since this is div element and they are block by default.

coding-challenges/pong-game/styles.css
```diff
@@ -0,0 +1,70 @@
...
+  background-color: #002e17;
+  color: #d3f9e4;
+  text-align: center;
+}
+
+.board {
+  border-radius: 14px;
+}
+
+.game-rules {
```
jdwilkin4, line 12: we should round out the corners here and remove the border color so it better matches the proposed designs in the issue. Quincy really wants these update projects to have a cleaner and more professional looking UI then they have in the past. Also some space between the title and the rules would be nice.

coding-challenges/pong-game/styles.css
```diff
@@ -0,0 +1,70 @@
...
+.inline {
+  justify-content: space-between;
+  display: flex;
+}
+
+.green-border {
+  border: 3px solid green;
+}
+
+button {
```
jdwilkin4, line 58: the buttons should better match the proposed designs on the issue to is looks more polished. Also a background color change on hover and cursor pointer would be nice too.

coding-challenges/pong-game/script.js
```diff
@@ -0,0 +1,208 @@
...
+let delta = 1;
+let lastMoveTime = 0; 
+let playerScore = 0;
+let computerScore = 0;
+const fps = 25;
+const playerOneScoreElement = document.querySelector("#player-score");
+const playerTwoScoreElement = document.querySelector("#computer-score");
+const playerOne = document.querySelector(".player_1");
+const canvas = document.getElementById("board");
+// @ts-ignore
```
jdwilkin4: do we need all of these ts-ignores? We should be fine to remove them.

coding-challenges/pong-game/script.js
```diff
@@ -0,0 +1,208 @@
...
+
+// Game Loop
+/**
+ * @param {number | undefined} [currentTime]
+ */
+function gameLoop(currentTime) {
+  if (lastTime !== undefined && currentTime !== undefined) {
+    // @ts-ignore
+    delta = Math.round(currentTime - lastTime);
+    console.log(`Frame time: ${Math.round(delta)}ms`);
```
jdwilkin4: looks like this was only here for testing. so it could be removed.

jdwilkin4 (commented):
I was able to test it out locally and play the game and here are my thoughts. 

While I wasn't able to beat the computer, I feel like the computer motion is still to much. We should replicate the behavior found in the original demo project here. The original slightly adjusted its position when necessary whereas here I felt here it was constantly moving. 

https://codepen.io/satyamdev/full/pdMmBp

Also we should have the ability for users to move the mouse to control the paddle like in the original. I wasn't able to do that here.

Another thing, the paddle should not be able to move completely off screen. In the original, you were able to move it partially off screen but not completely. 

Lastly, there should be some better messaging for when the game resets.  Right now the scores automatically reset when you click the reset button. But I feel like that could be easily missed. When I first clicked on the button I didn't notice that the game had reset but the paddle where still moving. A nicer user experience would be if there was some sort of modal that maybe asks if you are sure to reset the game. And if so, then they can start a new game or resume the one they have. 

Hope all of that helps 👍🏾

coding-challenges/pong-game/script.js
a2937: If  I don't have them, Visual Studio Code keeps giving me messages. I wanna remove them too but it'd be nice to not have all those messages.

coding-challenges/pong-game/styles.css
a2937: What color would be appropriate? In one of my later commits I set it to the light green used elsewhere in the game.

jdwilkin4: I was messing around with it locally and came up with these styles

```css
button {
  padding: 0.5rem 1rem;
  background-color: #16a34a;
  border: none;
  border-radius: 10px;
  cursor: pointer;
}

button:hover {
  background-color: #169247;
}
```

It differs from the original issue screenshot but it still looks clean IMO. 

Also, we should ditch the start button and reset button styles earlier up in the sheet and just have one set of styles to target all buttons

a2937:
> While I wasn't able to beat the computer, I feel like the computer motion is still to much. We should replicate the behavior found in the original demo project here. The original slightly adjusted its position when necessary whereas here I felt here it was constantly moving.

I'm still having trouble figuring out what you mean. I know there's a difference. But I see the computer paddle moving near constantly like mine. 


[Paddle movement](https://github.com/user-attachments/assets/21e60b7f-3ee4-41da-829f-f413ba7ba86a)

coding-challenges/pong-game/styles.css
a2937: I've added some space.

jdwilkin4:
Here is what I am seeing between the current demo and your project.

The current demo is smoother is more reactionary. Here is an example where I am completely still and the computer only moves slightly into position to beat me.It moves into place but its smooth. 

https://github.com/user-attachments/assets/99edc42e-7059-4af1-bb96-bb86d54dc2df


Whereas with yours, it is constantly moving up and down even when I am stationary. It seems more frantic than natural.


https://github.com/user-attachments/assets/a0bdc698-6190-4c22-ab17-086ded6b5b41

So we want to incorperate some of that smoothess the original had. I would suggest reviewing the source code to see how they got it to be smoother. 👍🏾 

Also, all of this will take a back seat for now until the TypeScript stuff is done 👍🏾

jdwilkin4 (approved):
I spent some time playing around with it and realized the ball speed was a little fast which made it seem like the paddles were moving frantically. So I slowed that down.

I also pushed up commits for the following:

- custom reset modal
- removed instances of var
- removed unused variables

Played the game for a while and I think it looks good. Thanks for all of your hard work on this. Let's merge it in 👍🏾

PR #933, opened 2025-07-20
feat: create typescript flashcard quiz workshop

### LINKED ISSUE #644: feat: create typescript flashcard quiz workshop

This will be lab no.2

a2937:
Checklist:

Closes #644

While not ready to merge at the time of this pull request (more CSS and Html changes definitely need to be made), I want to get more input regarding the ideas behind the workshop. 

The purpose I was aiming towards was focused on DOM manipulation and how that can be handled with Typescript. I know there's more I should be doing with this; something more in-depth.

### DIFF
--- a/fullstack-cert/typescript-projects/flashcard-quiz-app/src/index.html
+++ b/fullstack-cert/typescript-projects/flashcard-quiz-app/src/index.html
@@ -18,10 +18,10 @@
     <hr>
     <form class="entry-form">
       <label for="front-text">Front:</label>
-      <input id="front-text"/>
+      <input id="front-text" required/>
       <br/>
       <label for="back-text">Back:</label>
-      <input id="back-text"/> 
+      <input id="back-text" required/> 
       <br/>
       <p id="entry-error" class="error"></p>
       <button onclick="uploadNewCard()">Upload new flash card</button>


jdwilkin4:
Is there a reason why we are not using React with TypeScript?
Are all of these workshops going to be with vanilla JS and TypeScript? If so, we should strongly consider adding additional workshops and labs with React and TypeScript since campers will need that practice to get familiar with real world apps.

jdwilkin4 (commented):
Tested in locally and I think it is off to a good start but it could use some more TypeScript IMO.

Two things stuck out to me.


We should have some sort of custom error message when a user doesn't supply text for front or back. We should use a traditional form element and basic HTML validation for the inputs. But it would be nice to have a custom error message and that will give campers more practice with TypeScript. 

Also, it would be nice to have a little bit of keyboard functionality.  Right now there is a flip function. But it would be nice to be bale to flip a card with either a click on the mouse or even a spacebar or enter key. And that will give more practice working with the `KeyboardEvent` interface.

a2937:
> Is there a reason why we are not using React with TypeScript? Are all of these workshops going to be with vanilla JS and TypeScript? If so, we should strongly consider adding additional workshops and labs with React and TypeScript since campers will need that practice to get familiar with real world apps.

I was hoping to introduce React and Typescript in Workshop 3 and then have it be used in the Lab. Workshop 1 is going to be focused on fetching data from an external source. The main app's purpose will be a fortune-telling app. I'm fairly certain I picked out a Tarot deck.

PR #949, opened 2025-08-04
feat: add flashcard app workshop

### LINKED ISSUE #644: feat: create typescript flashcard quiz workshop

naomi-lgbt:
Checklist:

- [ ] I have read [freeCodeCamp's contribution guidelines](https://contribute.freecodecamp.org).
- [ ] My pull request has a [descriptive title](https://contribute.freecodecamp.org/#/how-to-open-a-pull-request?id=prepare-a-good-pr-title) (**not** a vague title like `Update index.md`)

Closes #644 
Replaces #933 

## List of TypeScript Concepts taught

- Type Annotations
- Interfaces
- Type Guards
- Generic DOM Typing
- Union & Type Narrowing
- Type-safe error handling 
- Typed Events


eastbetsy:
**Complete the flashcard quiz app with a cleaner UI and control panel to make it look like an actual app.**
Credits to Musa for the Typescript. I just fixed it a bit when I edited the HTML and CSS for the new UI.
Not sure what to do for the quiz part. I hope this passes for now. Let me know if anything needs fixed. Sorry for the inactivity!

shootermv:
i think it can be better to convert all the separate methods (like `refresh` or `flipcard`), the code has currently - to private methods of FlashCardGame class...
it will more `typescript-style`

fullstack-cert/typescript-projects/flashcard-quiz-app/src/script.ts
```diff
@@ -0,0 +1,226 @@
...
+      wrongCount: 0,
+      gameStarted: false,
+      gameCompleted: false,
+      currentDeck: [...cardDecks.general.cards],
+    };
+
+    this.elements = {
+      flashcard: document.querySelector(".flashcard")!,
+      cardsList: document.getElementById("cards-list")!,
+      flipBtn: document.getElementById("flip-btn")!,
```
naomi-lgbt: Rather than using non-null assertions here, we should have the campers implement a fallback approach if the elements are not found.

Maybe something like injecting an element with text "DOM failed to hydrate, refresh the page"?

fullstack-cert/typescript-projects/flashcard-quiz-app/src/script.ts
```diff
@@ -0,0 +1,226 @@
+const cardDisplay = document.querySelector("#current-card") as HTMLElement;
+const cardButtonsContainer = document.getElementById(
+  "cards-list",
+) as HTMLElement;
+
+const frontInput = document.getElementById("front-text") as HTMLTextAreaElement;
+const backInput = document.getElementById("back-text") as HTMLTextAreaElement;
```
naomi-lgbt:
```suggestion
const cardDisplay = document.querySelector<HTMLElement>("#current-card");
const cardButtonsContainer = document.getElementById<HTMLElement>(
  "cards-list",
);

const frontInput = document.getElementById<HTMLTextAreaElement>("front-text");
const backInput = document.getElementById<HTMLTextAreaElement>("back-text");
```

Rather than type casting, we should teach campers to pass the generic.

fullstack-cert/typescript-projects/flashcard-quiz-app/src/script.ts
```diff
@@ -0,0 +1,226 @@
...
+  backText: string;
+}
+
+class InvalidUserInputError extends Error {
+  constructor(message: string) {
+    super(message);
+    this.name = "InvalidUserInputError";
+    Object.setPrototypeOf(this, InvalidUserInputError.prototype);
+  }
+}
```
naomi-lgbt, line 21: Creating a custom error class *might* be overkill. Will need to think on this.

fullstack-cert/typescript-projects/flashcard-quiz-app/src/script.ts
```diff
@@ -0,0 +1,226 @@
...
+      "No card selected. Add a new card to get started!";
+    return;
+  }
+
+  const card = currentCards[currentCardIndex];
+  cardDisplay.querySelector(".card-front").textContent = card.frontText;
+  cardDisplay.querySelector(".card-back").textContent = card.backText;
+  // add correct background to current card
+  Array.from(cardButtonsContainer.children).forEach((child, i) => {
+    const button = child as HTMLElement;
```
naomi-lgbt: We could maybe turn this into a type guard? 

```ts
const isButtonElement = (element: unknown): element is HTMLButtonElement => {
  return element instanceof HTMLButtonElement;
}
```

Something like this - have not tested the `instanceof`.

fullstack-cert/typescript-projects/flashcard-quiz-app/src/script.ts
```diff
@@ -0,0 +1,226 @@
...
+    return;
+  }
+
+  currentCards.splice(currentCardIndex, 1);
+  cardButtonsContainer.removeChild(
+    cardButtonsContainer.children[currentCardIndex],
+  );
+
+  Array.from(cardButtonsContainer.children).forEach((child, i) => {
+    (child as HTMLButtonElement).onclick = () => {
```
naomi-lgbt: Then we could use the typeguard here too.

fullstack-cert/typescript-projects/flashcard-quiz-app/src/script.ts
```diff
@@ -0,0 +1,226 @@
...
+      currentCardIndex = i;
+      showingFront = true;
+      refresh();
+    };
+  });
+
+  if (currentCards.length === 0) {
+    currentCardIndex = -1;
+  } else {
+    currentCardIndex = Math.max(0, currentCardIndex - 1);
```
naomi-lgbt: Should we also clamp this to the length of the array?

fullstack-cert/typescript-projects/flashcard-quiz-app/src/script.ts
```diff
@@ -0,0 +1,226 @@
...
+    showingFront = true;
+    refresh();
+  };
+  return btn;
+}
+
+function uploadNewCard(): void {
+  const errorElement = document.getElementById(
+    "entry-error",
+  ) as HTMLParagraphElement;
```
naomi-lgbt:
```suggestion
  const errorElement = document.getElementById<HTMLParagraphElement>(
    "entry-error",
  );
```

fullstack-cert/typescript-projects/flashcard-quiz-app/src/script.ts
```diff
@@ -0,0 +1,226 @@
...
+      console.error("An unexpected error occurred:", ex);
+    }
+  }
+}
+
+document.addEventListener("keydown", (ev: KeyboardEvent) => {
+  const frontInput = document.getElementById(
+    "front-text",
+  ) as HTMLTextAreaElement;
+  const backInput = document.getElementById("back-text") as HTMLTextAreaElement;
```
naomi-lgbt:
```suggestion
  const frontInput = document.getElementById<HTMLTextAreaElement>(
    "front-text",
  );
  const backInput = document.getElementById<HTMLTextAreaElement>("back-text");
```

a2937:
Wait. I think I made a mistake by applying those suggestions. I need to update the Typescript version to make it recognizes `querySelector` and `getElementById` as generic.

shootermv:
i have ui suggestion: to move `flip` button to be below the card
and rename the right section to `manage cards`, like here:
image
Also instead of `back` & `front` - will it be better to change it to `question` & `answer`? 
what do you think? @ElectricMolasses @a2937 ?

fullstack-cert/typescript-projects/flashcard-quiz-app/src/script.ts
Rhombert: :heavy_check_mark:

fullstack-cert/typescript-projects/flashcard-quiz-app/src/script.ts
Rhombert: :heavy_check_mark:

a2937:
Sure.

shootermv:
one more question: do we need `typescript` package listed in the package.json list?
currently it is missing

Rhombert:
> i have ui suggestion: to move `flip` button to be below the card and rename the right section to `manage cards`, like here: image Also instead of `back` & `front` - will it be better to change it to `question` & `answer`? what do you think? @ElectricMolasses @a2937 ?

It's hard to beat back and front semantically. It immediately tells me that there is a back and front face for this css, and explains why there are properties like `RotateY`. `question` and `answer` describes none of the implementation to me, only what the text is going to represent.

If we do see a benefit in question and answer, I would prefer those be added to the text that the front and back of the cards contain, rather than the card faces themselves, or even as an additional class to the cards. `.front` and `.back` shouldn't be removed/replaced.

shootermv:
it can be better to move all the methods to be a class-member methods instead

a2937:
@shootermv Why did you remove the keydown handler?

shootermv:
> @shootermv Why did you remove the keydown handler?

i removed it since it wasnt doing noting (i suppose the initial intention was - to flip the card when user presses enter, which looks not so correct to me)

fullstack-cert/typescript-projects/flashcard-quiz-app/src/script.ts
```diff
@@ -0,0 +1,244 @@
...
+    cardButtonsContainer.children[currentCardIndex],
+  );
+
+  Array.from(cardButtonsContainer.children).forEach((child, i) => {
+    if (!isButtonElement(child)) {
+      console.warn(`Element {${child}} is not a button.`);
+      return;
+    }
+
+    (child as HTMLButtonElement).onclick = () => {
```
naomi-lgbt:
```suggestion
    child.onclick = () => {
```

The typeguard above makes this cast unnecessary.

fullstack-cert/typescript-projects/flashcard-quiz-app/src/script.ts
```diff
@@ -0,0 +1,244 @@
...
+    if (!isButtonElement(child)) {
+      console.warn(`Element {${child}} is not a button.`);
+      return;
+    }
+    const button = child as HTMLElement;
+    if (i === currentCardIndex) {
+      button.classList.add("selected");
+    } else {
+      button.classList.remove("selected");
+    }
```
naomi-lgbt:
```suggestion
    if (i === currentCardIndex) {
      child.classList.add("selected");
    } else {
      child.classList.remove("selected");
    }
```

The typeguard means we do not need to cast, so we can remove this extraneous variable too.

shootermv:
im little concerned by the fact that there is some methods class based and others - not...

PR #951, opened 2025-08-04
feat: add resume builder workshop

### LINKED ISSUE #645: feat: create typescript resume builder workshop

This will be lab no.3

naomi-lgbt:
Checklist:

- [ ] I have read [freeCodeCamp's contribution guidelines](https://contribute.freecodecamp.org).
- [ ] My pull request has a [descriptive title](https://contribute.freecodecamp.org/#/how-to-open-a-pull-request?id=prepare-a-good-pr-title) (**not** a vague title like `Update index.md`)

Closes #645
Replaces #943


shootermv:
i can see `form` closing tag is missing at index.html...

shootermv:
here is my suggestion for UI:
the app will have `preview` and `edit` modes:
### `preview`  mode:
image     
     
at `preview` mode - all user details will display,
at `edit` - you can be able to insert or edit the details

### `edit` mode:
image
    
================================================================
DATA STRUCTURE:
cv data should consist of: 
- personal details [Object](email, address, fullname, phone)
- work experiences  [Array](company, position, startDate, endDate)
- educations [Array](institution, degree, startDate, endDate)
- skills [Array](name, level)

fullstack-cert/typescript-projects/resume-builder/src/Field.tsx
naomi-lgbt, line 1: This will need to be folded into `App.tsx`, our editor does not currently support multiple script files.

fullstack-cert/typescript-projects/resume-builder/src/App.tsx
```diff
@@ -0,0 +1,131 @@
+import { FileText, User } from "lucide-react";
```
naomi-lgbt: We will need to fetch this from something like CDN.js in the HTML to make it available in the code.

fullstack-cert/typescript-projects/resume-builder/src/App.tsx
```diff
@@ -0,0 +1,131 @@
...
+    setPhone((e.target as HTMLInputElement).value);
+  }
+
+  function handleUpdateObjective(e: ChangeEvent) {
+    setObjective((e.target as HTMLInputElement).value);
+  }
+
+  function handleUpdateDegree(e: ChangeEvent) {
+    setDegree((e.target as HTMLInputElement).value);
+  }
```
naomi-lgbt: Typeguard here, rather than casting.

fullstack-cert/typescript-projects/resume-builder/src/App.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+  function handleUpdateGradYear(e: ChangeEvent) {
+    const gradYear = parseInt((e.target as HTMLInputElement).value);
+    const cutOffYear = 1976;
+    if (Number.isSafeInteger(gradYear) && gradYear > cutOffYear) {
+      setGradYear(parseInt((e.target as HTMLInputElement).value));
+    } else {
```
naomi-lgbt: Same here.

fullstack-cert/typescript-projects/resume-builder/src/components/CVPreview.tsx
naomi-lgbt, line 1: All of these components will need to be folded into the App.tsx too.

fullstack-cert/typescript-projects/resume-builder/src/App.tsx
a2937: These don't look used anywhere?

shootermv:
question: why not to use `lucide-react` library? (it looks for my much complicate to use imported icons...)

a2937:
Because I couldn't figure out how to add it with CDN.js AND get all the types to work.

shootermv:
> Because I couldn't figure out how to add it with CDN.js AND get all the types to work.


let me understand, you cannot use `npm` packages at all at those projects? instead you plan to import all the libs like react, by using cdn.js?

Rhombert:
> > Because I couldn't figure out how to add it with CDN.js AND get all the types to work.
> 
> let me understand, you cannot use `npm` packages at all at those projects? instead you plan to import all the libs like react, by using cdn.js?

There're grounds for FCC to take this approach. Under the assumption that they use this approach for all of their external libraries (I don't know that they do, just justifying the extreme case), there have been multiple incidents in the past where npm has been a delivery device for exploits, across various packages. By storing approved versions of packages on your own CDN you help to protect against this form of attack.

It also protects from the pettier issues, like when a particular javascript library was removed from npm and broke all dependent libraries. When stored on an internal CDN you know the version you're using is guaranteed to stick around. The `left-pad` incident, if you want to read about it yourself.

naomi-lgbt:
Let's scrap lucide and use font awesome instead - those are readily available at https://cdnjs.com/libraries/font-awesome

PR #969, opened 2025-08-21
chore: added prototype for classes and objects lab

### LINKED ISSUE #953: Create a prototype for classes and objects lab
This would be a quick lab to practice classes and objects in Python. We could also consider to reuse an existing project.

fhsinchy:
Checklist:

Closes #953

### DIFF
--- a/fullstack-cert/python-projects/classes-and-objects/main.py
+++ b/fullstack-cert/python-projects/classes-and-objects/main.py
@@ -15,16 +15,15 @@ class Planet:
         
 
 # Test the Planet class
-if __name__ == "__main__":
-    planet_1 = Planet('Earth', 'terrestrial', 'Sun')
-    planet_2 = Planet('Jupiter', 'gas giant', 'Sun')
-    planet_3 = Planet('Kepler-452b', 'super-Earth', 'Kepler-452')
+planet_1 = Planet('Earth', 'terrestrial', 'Sun')
+planet_2 = Planet('Jupiter', 'gas giant', 'Sun')
+planet_3 = Planet('Kepler-452b', 'super-Earth', 'Kepler-452')
 
-    print(planet_1.orbit()) # Earth is orbiting around Sun...
-    print(planet_1) # Planet: Earth | Type: terrestrial | Star: Sun
+print(planet_1.orbit()) # Earth is orbiting around Sun...
+print(planet_1) # Planet: Earth | Type: terrestrial | Star: Sun
 
-    print(planet_2.orbit()) # Jupiter is orbiting around Sun...
-    print(planet_2) # Planet: Jupiter | Type: gas giant | Star: Sun
+print(planet_2.orbit()) # Jupiter is orbiting around Sun...
+print(planet_2) # Planet: Jupiter | Type: gas giant | Star: Sun
 
-    print(planet_3.orbit()) # Kepler-452b is orbiting around Kepler-452...
-    print(planet_3) # Planet: Kepler-452b | Type: super-Earth | Star: Kepler-452
+print(planet_3.orbit()) # Kepler-452b is orbiting around Kepler-452...
+print(planet_3) # Planet: Kepler-452b | Type: super-Earth | Star: Kepler-452
--- a/fullstack-cert/python-projects/classes-and-objects/main.py
+++ b/fullstack-cert/python-projects/classes-and-objects/main.py
@@ -2,9 +2,9 @@ class Planet:
     def __init__(self, name, planet_type, star):
         args = (name, planet_type, star)
         if not all(isinstance(arg, str) for arg in args):
-            raise ValueError("Error: Name, planet type, and star must be strings")
+            raise ValueError("name, planet type, and star must be strings")
         if not all(args):
-            raise ValueError('name, planet_type, and star should be non-empty strings')
+            raise ValueError('name, planet_type, and star must be non-empty strings')
         self.name = name
         self.planet_type = planet_type
         self.star = star   # The star the planet orbits
--- a/fullstack-cert/python-projects/classes-and-objects/main.py
+++ b/fullstack-cert/python-projects/classes-and-objects/main.py
@@ -1,61 +1,30 @@
-class BankAccount:
-    def __init__(self, account_holder, initial_balance=0):
-        self.__account_holder = account_holder
-        self.__balance = initial_balance
-        self.__transaction_history = []
+class Planet:
+    def __init__(self, name, planet_type, star):
+        if not name or not planet_type or not star:
+            print("Error: Name, planet type, and star cannot be empty")
+            return
+        self.name = name
+        self.planet_type = planet_type
+        self.star = star   # The star the planet orbits
 
-    def deposit(self, amount):
-        if amount > 0:
-            self.__balance += amount
-            self.__transaction_history.append(f"Deposited ${amount}")
-            return f"Deposited ${amount}. New balance: ${self.__balance}"
-        else:
-            return "Deposit amount must be positive"
-
-    def withdraw(self, amount):
-        if amount > 0:
-            if amount <= self.__balance:
-                self.__balance -= amount
-                self.__transaction_history.append(f"Withdrew ${amount}")
-                return f"Withdrew ${amount}. New balance: ${self.__balance}"
-            else:
-                return "Insufficient funds"
-        else:
-            return "Withdrawal amount must be positive"
-
-    def get_balance(self):
-        return self.__balance
-
-    def get_account_holder(self):
-        return self.__account_holder
-
-    def get_transaction_history(self):
-        return self.__transaction_history.copy()
+    def orbit(self):
+        return f'{self.name} is orbiting around {self.star}...'
 
     def __str__(self):
-        return f"Account Holder: {self.__account_holder}, Balance: ${self.__balance}"
+        return f'Planet: {self.name} | Type: {self.planet_type} | Star: {self.star}'
+        
 
-
-# Test the BankAccount class
+# Test the Planet class
 if __name__ == "__main__":
-    # Create a new bank account
-    account = BankAccount("John Doe", 1000)
-    
-    # Test deposit
-    print(account.deposit(500))
-    
-    # Test withdraw
-    print(account.withdraw(200))
-    
-    # Test insufficient funds
-    print(account.withdraw(2000))
-    
-    # Test invalid amounts
-    print(account.deposit(-100))
-    print(account.withdraw(-50))
-    
-    # Display account info
-    print(account)
-    print(f"Account holder: {account.get_account_holder()}")
-    print(f"Current balance: ${account.get_balance()}")
-    print(f"Transaction history: {account.get_transaction_history()}")
+    planet_1 = Planet('Earth', 'terrestrial', 'Sun')
+    planet_2 = Planet('Jupiter', 'gas giant', 'Sun')
+    planet_3 = Planet('Kepler-452b', 'super-Earth', 'Kepler-452')
+
+    print(planet_1.orbit()) # Earth is orbiting around Sun...
+    print(planet_1) # Planet: Earth | Type: terrestrial | Star: Sun
+
+    print(planet_2.orbit()) # Jupiter is orbiting around Sun...
+    print(planet_2) # Planet: Jupiter | Type: gas giant | Star: Sun
+
+    print(planet_3.orbit()) # Kepler-452b is orbiting around Kepler-452...
+    print(planet_3) # Planet: Kepler-452b | Type: super-Earth | Star: Kepler-452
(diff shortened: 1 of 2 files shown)


fhsinchy:
@Dario-DC, @zairahira I think I went a bit overboard with this one but most of the code is repetitive so can be shortened. Implemented a library management system with multiple classes with properties and methods.

Dario-DC:
@fhsinchy  We need something much simpler...
This lab would come right after https://github.com/freeCodeCamp/freeCodeCamp/pull/61750

So just one class with two or three methods including an `__init__` is enough.

fhsinchy:
@Dario-DC I have simplified the code according to your suggestion.

fullstack-cert/python-projects/classes-and-objects/user-stories.md
```diff
@@ -0,0 +1,10 @@
+1. Create a class named `BankAccount` with an `__init__` method that takes `account_holder` and `initial_balance` (default 0) parameters.
+2. In the `__init__` method, create private attributes `__account_holder`, `__balance`, and `__transaction_history` (empty list).
```
majestic-owl448:
```suggestion
2. In the `__init__` method, create attributes `account_holder`, `balance`, and `transaction_history` (empty list).
```

topic unavaible at this point

Dario-DC:
Another possible idea:
```py
class Planet:
    def __init__(self, name, planet_type, star):
        self.name = name
        self.planet_type = planet_type
        self.star = star   # The star the planet orbits

    def orbit(self):
        print(f'{self.name} is orbiting around {self.star}...')

    def __str__(self):
        return f'Planet: {self.name} | Type: {self.planet_type} | Star: {self.star}'
        

planet_1 = Planet('Earth', 'terrestrial', 'Sun')
planet_2 = Planet('Jupiter', 'gas giant', 'Sun')
planet_3 = Planet('Kepler-452b', 'super-Earth', 'Kepler-452')


planet_1.orbit() # Earth is orbiting around Sun...
print(planet_1) # Planet: Earth | Type: terrestrial | Star: Sun

planet_2.orbit() # Jupiter is orbiting around Sun...
print(planet_2) # Planet: Jupiter | Type: gas giant | Star: Sun

planet_3.orbit() # Kepler-452b is orbiting around Kepler-452...
print(planet_3) # Planet: Kepler-452b | Type: super-Earth | Star: Kepler-452
```
If we do not want to use `__str__` we can replace it with something like:
```py
def get_info(self):
    return f'Planet: {self.name} | Type: {self.planet_type} | Star: {self.star}'
```

fhsinchy:
The example by @Dario-DC looks good. I can rework it into something else if it's too similar to the budget app, I can also get rid of the unavailable topics @ilenia-magoni

Dario-DC:
If we want to avoid having similar projects next to each other, then I think we can use this.

> Another possible idea:
> 
> ```python
> class Planet:
>     def __init__(self, name, planet_type, star):
>         self.name = name
>         self.planet_type = planet_type
>         self.star = star   # The star the planet orbits
> 
>     def orbit(self):
>         print(f'{self.name} is orbiting around {self.star}...')
> 
>     def __str__(self):
>         return f'Planet: {self.name} | Type: {self.planet_type} | Star: {self.star}'
>         
> 
> planet_1 = Planet('Earth', 'terrestrial', 'Sun')
> planet_2 = Planet('Jupiter', 'gas giant', 'Sun')
> planet_3 = Planet('Kepler-452b', 'super-Earth', 'Kepler-452')
> 
> 
> planet_1.orbit() # Earth is orbiting around Sun...
> print(planet_1) # Planet: Earth | Type: terrestrial | Star: Sun
> 
> planet_2.orbit() # Jupiter is orbiting around Sun...
> print(planet_2) # Planet: Jupiter | Type: gas giant | Star: Sun
> 
> planet_3.orbit() # Kepler-452b is orbiting around Kepler-452...
> print(planet_3) # Planet: Kepler-452b | Type: super-Earth | Star: Kepler-452
> ```
> 
> If we do not want to use `__str__` we can replace it with something like:
> 
> ```python
> def get_info(self):
>     return f'Planet: {self.name} | Type: {self.planet_type} | Star: {self.star}'
> ```

fhsinchy:
@Dario-DC yeah I'm using yours as a base and seeing if I can add anything to it.

fhsinchy:
@Dario-DC, @ilenia-magoni I have updated the project with the planetary code with some minor changes.

fullstack-cert/python-projects/classes-and-objects/user-stories.md
```diff
@@ -0,0 +1,8 @@
+1. Create a class named `Planet` with an `__init__` method that takes `name`, `planet_type`, and `star` parameters.
+2. In the `__init__` method, add validation to check that none of the parameters are empty strings, and print an error message if any are empty.
+3. In the `__init__` method, create instance attributes `self.name`, `self.planet_type`, and `self.star` to store the parameter values.
+4. Create an `orbit` method that returns a string in the format `'{name} is orbiting around {star}...'`.
+5. Create a `__str__` method that returns a string in the format `'Planet: {name} | Type: {planet_type} | Star: {star}'`.
+6. Test your class by creating three Planet instances with different names, types, and stars.
+7. Call the `orbit` method on each planet and print the returned string.
+8. Print each planet object to see the `__str__` method output.
```
Dario-DC:
```suggestion
1. You should create a class named `Planet`.
1. The `Planet` class should have an `__init__` method that:
   - Has four parameters: `self`, `name`, `planet_type`, and `star`.
   - Prints the message `Error: name, planet type, and star should be non-empty strings` if any of the arguments passed in is either a non-string type or an empty string.
   - Assigns the values passed in to the instance attributes `name`, `planet_type`, and `star`.
1. The `Planet` class should have an `orbit` method that returns a string in the format `{name} is orbiting around {star}...`.
1. The `Planet` class should have a `__str__` method that returns a string in the format `Planet: {name} | Type: {planet_type} | Star: {star}`.
1. You should create three instances of the `Planet` class named `planet_1`, `planet_2`, and `planet_3`.
1. You should print each planet object to see the `__str__` method output.
1. You should call the `orbit` method on each planet object and print the result.
```

fullstack-cert/python-projects/classes-and-objects/main.py
```diff
@@ -0,0 +1,30 @@
+class Planet:
+    def __init__(self, name, planet_type, star):
+        if not name or not planet_type or not star:
+            print("Error: Name, planet type, and star cannot be empty")
```
Dario-DC:
```suggestion
       for arg in (name, planet_type, star):
            if not isinstance(arg, str) or not arg:
                print("Error: name, planet type, and star should be non-empty strings")
                return
```

fullstack-cert/python-projects/classes-and-objects/main.py
```diff
@@ -0,0 +1,30 @@
...
+
+    def orbit(self):
+        return f'{self.name} is orbiting around {self.star}...'
+
+    def __str__(self):
+        return f'Planet: {self.name} | Type: {self.planet_type} | Star: {self.star}'
+        
+
+# Test the Planet class
+if __name__ == "__main__":
```
Dario-DC: This `if` should be required or removed. Otherwise it's difficult to verify what they are doing.

Dario-DC (commented):
Left a few comments to make user stories more consistent with what we've done so far and improve validation.

fullstack-cert/python-projects/classes-and-objects/main.py
fhsinchy: I think we can get rid of it for simplicity

fhsinchy:
@Dario-DC final review

fullstack-cert/python-projects/classes-and-objects/main.py
```diff
@@ -0,0 +1,29 @@
+class Planet:
+    def __init__(self, name, planet_type, star):
+        if not name or not planet_type or not star:
+            print("Error: Name, planet type, and star cannot be empty")
```
gikf:
```suggestion
            raise ValueError("Error: Name, planet type, and star cannot be empty")
```
It would be better to raise exception here.

Otherwise object is still created, but not fully initialized. Which means there can be _two_ `Planet` objects - fully initialized and not. Trying to use the latter in a normal way would cause exception.

fullstack-cert/python-projects/classes-and-objects/main.py
```diff
@@ -0,0 +1,29 @@
+class Planet:
+    def __init__(self, name, planet_type, star):
+        if not name or not planet_type or not star:
```
gikf:
```suggestion
        if not all((name, planet_type, star)):
```

fullstack-cert/python-projects/classes-and-objects/main.py
fhsinchy: I had this on the original prototype but left it off due to how little the camper has learned at this point. It was done for simplification.

fullstack-cert/python-projects/classes-and-objects/main.py
fhsinchy: Same goes for the usage of the `all()` function.

fullstack-cert/python-projects/classes-and-objects/main.py
Dario-DC: Actually this is good point and we can use `raise` since it has been already taught.

fullstack-cert/python-projects/classes-and-objects/main.py
Dario-DC: This is not relevant because we are going to test the result, not the implementation. But it needs to be updated because the current user story says:
> Prints the message `Error: name, planet type, and star should be non-empty strings` if any of the arguments passed in is either **a non-string type** or an empty string.`

fhsinchy: @Dario-DC what do you suggest on this then? Should we go with the usage of `all()`?

Dario-DC: The use of `all` doesn't matter here because we are not requiring it. Campers are free to write the `if` condition in many ways. The thing is that `Planet('Earth', 'terrestrial', 0)` should raise an error. `0` is not a string though. So this should be a `TypeError` actually.

So I would go with:
```py
class Planet:
    def __init__(self, name, planet_type, star):
        args = (name, planet_type, star)
        if not all(isinstance(arg, str) for arg in args):
            raise TypeError('name, planet_type, and star should be strings')
            return
        if not all(args):
            raise ValueError('name, planet_type, and star should be non-empty strings')
            return
    ...
```
And update the user stories:

```diff
-   - Prints the message `Error: name, planet type, and star should be non-empty strings` if any of the arguments passed in is either a non-string type or an empty string.
+   - Raises a `TypeError` with the message `name, planet type, and star should be strings` if any of the arguments passed in a non-string type.
+   - Raises a `ValueError` with the message `name, planet type, and star should be non-empty strings` if any of the arguments passed in is an empty string.
```

fhsinchy:
@Dario-DC I've replaced the prints with raises and updated the stories accordingly.

fullstack-cert/python-projects/classes-and-objects/main.py
```diff
@@ -0,0 +1,31 @@
+class Planet:
+    def __init__(self, name, planet_type, star):
+        args = (name, planet_type, star)
+        if not all(isinstance(arg, str) for arg in args):
+            raise ValueError("Error: Name, planet type, and star must be strings")
```
Dario-DC:
```suggestion
            raise ValueError("name, planet type, and star must be strings")
```

fullstack-cert/python-projects/classes-and-objects/main.py
```diff
@@ -0,0 +1,31 @@
...
+        if not all(args):
+            raise ValueError('name, planet_type, and star should be non-empty strings')
```
Dario-DC:
```suggestion
            raise ValueError('name, planet_type, and star must be non-empty strings')
```

fullstack-cert/python-projects/classes-and-objects/user-stories.md
```diff
@@ -0,0 +1,11 @@
+1. You should create a class named `Planet`.
+1. The `Planet` class should have an `__init__` method that:
+   - Has four parameters: `self`, `name`, `planet_type`, and `star`.
+   - Raises a `ValueError` with the message `Error: Name, planet type, and star must be strings` if any of the arguments passed in is not a string type.
```
Dario-DC:
```suggestion
   - Raises a `ValueError` with the message `name, planet type, and star must be strings` if any of the arguments passed in is not a string type.
```

fullstack-cert/python-projects/classes-and-objects/user-stories.md
```diff
@@ -0,0 +1,11 @@
...
+   - Raises a `ValueError` with the message `name, planet_type, and star should be non-empty strings` if any of the arguments passed in is an empty string.
```
Dario-DC:
```suggestion
   - Raises a `ValueError` with the message `name, planet_type, and star must be non-empty strings` if any of the arguments passed in is an empty string.
```

Dario-DC (commented):
Updated error messages for consistency. Then, let's merge this

PR #973, opened 2025-08-29
feature selection for chekbox lab project

### LINKED ISSUE #778: Create styling checkboxes lab prototype
We have this issue here for creating the workshop that will go before this 

https://github.com/freeCodeCamp/CurriculumExpansion/issues/777

This issue is about create the lab prototype for styling checkboxes

Ksound22:
Checklist:

Closes #778

### DIFF
--- a/fullstack-cert/css-projects/feature-selection/styles.css
+++ b/fullstack-cert/css-projects/feature-selection/styles.css
@@ -33,17 +33,34 @@ body {
   box-shadow: 0 4px 12px rgba(255, 201, 53, 0.6);
 }
 
-.feature-card input[type='checkbox'] {
+.feature-card input[type="checkbox"] {
   position: absolute;
   top: 15px;
   right: 15px;
   width: 24px;
   height: 24px;
   cursor: pointer;
+  appearance: none;
+  -webkit-appearance: none;
+  border: 2px solid #ccc;
+  border-radius: 4px;
+  background-color: white;
+  transition: all 0.3s;
 }
 
-.feature-card input[type='checkbox']:checked {
-  accent-color: #f1be32;
+.feature-card input[type="checkbox"]:checked {
+  background-color: #f1be32;
+  border-color: #f1be32;
+}
+
+.feature-card input[type="checkbox"]:checked::after {
+  content: "✓";
+  display: block;
+  text-align: center;
+  font-size: 16px;
+  font-weight: bold;
+  color: white;
+  line-height: 20px;
 }
 
 .feature-card h4 {
--- a/fullstack-cert/css-projects/feature-selection/styles.css
+++ b/fullstack-cert/css-projects/feature-selection/styles.css
@@ -1,0 +1,59 @@
+body {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  flex-direction: column;
+  height: 100vh;
+}
+
+.feature-card-container {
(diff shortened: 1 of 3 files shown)


majestic-owl448:
have we changed things and do not include the user stories with the prototype of labs anymore?

fullstack-cert/css-projects/feature-selection/styles.css
```diff
@@ -0,0 +1,59 @@
...
+  position: absolute;
+  top: 15px;
+  right: 15px;
+  width: 24px;
+  height: 24px;
+  cursor: pointer;
+}
+
+.feature-card input[type='checkbox']:checked {
+  /* Point out that accent-color changes the backfround of the check and changes the checkmark to black or white irrespective of the value applied. it's probably the property that bring something unique to styling checkboxes */
```
majestic-owl448: how would you point this out in a lab?

Ksound22:
> have we changed things and do not include the user stories with the prototype of labs anymore?

I see. I will add the user stories.

jdwilkin4:
I pushed up two commits to resolve two issues to help speed up the review process.

Issue no.1

Accent-color has limited browser availability. So we should avoid using it here 

https://developer.mozilla.org/en-US/docs/Web/CSS/accent-color

So I updated the code to remove it. 

Issue no.2

The user stories you added were to build out the exam demo project exactly. But remember that labs are different than workshops. Workshops are step based projects where the camper has to build out the code we tell them through steps. Labs are open ended projects with just an editor and user stories. The demo project you built will only be shown as an example for campers. But they will need to build out their own project different from the demo. Here is an example 
https://www.freecodecamp.org/learn/full-stack-developer/lab-page-of-playing-cards/build-a-page-of-playing-cards

I updated the user stories so they are opened ended. You can view that in my last commit. 👍🏾

fullstack-cert/css-projects/feature-selection/user-stories.md
```diff
@@ -0,0 +1,11 @@
+# User Stories
+
+1. Your page should have an `h1` element with the text `Feature Selection`.
+2. Your page should have a `div` element with the class `feature-card-container`.
```
jdwilkin4, line 4: after reviewing this, I think it is fine to remove this user story. It should be optional if they have a container for these label elements.

fullstack-cert/css-projects/feature-selection/styles.css
```diff
@@ -0,0 +1,74 @@
...
+  cursor: pointer;
+  transition: 0.3s;
+}
+
+.feature-card:hover {
+  border-color: #f1be32;
+  box-shadow: 0 4px 12px rgba(255, 201, 53, 0.6);
+}
+
+.feature-card input[type="checkbox"] {
```
jdwilkin4: the positioning of the checkboxes will need to be updated for mobile.
This is what I am currently seeing for mobile
Screenshot 2025-09-07 at 9 51 01 AM

jdwilkin4: it looks like setting both the top and right properties to something like 5px should work for all devices. but please test that on your end too.

Ksound22: Yes that works

jdwilkin4 (approved):

PR #986, opened 2025-09-10
feat(curriculum): add TypeScript prototype for generics workshop

### LINKED ISSUE #978: Create small workshop prototype for teaching generics and type guards/narrowing
We already have three larger TypeScript workshops. But we do need a couple of smaller ones so campers can just focus on a few key TypeScript concepts before jumping into the larger projects.

So we need a small workshop to teach generics and type guards/narrowing.
This workshop should eventually boil down to like 30 or less steps. For reference, the other TypeScript projects will probably hit closer to 100+ steps. So this project needs to be much smaller than the others and much simpler. This will go in the very beginning of the workshops.

You can come up with whatever UI and colors you like.

This one should also just use HTML, TypeScript and CSS. Not React.

ElectricMolasses:
Checklist:

Closes #978 

This is the initial pass that I believe is ready for feedback. I'm fairly confident I can break this into ~30 steps, in a similar fashion to existing live curriculum, but it's dependent on scaffolding in an initial state for the camper to begin work on, to keep their own code scoped down to what's directly applicable to the lesson. It can be trimmed down further without much work, but the current approach is what I personally view as ideal.

### DIFF
--- a/fullstack-cert/typescript-projects/generics-workshop/src/script.ts
+++ b/fullstack-cert/typescript-projects/generics-workshop/src/script.ts
@@ -2,10 +2,10 @@ abstract class Bug<T> {
   emoji!: T;
   emojiElement: HTMLParagraphElement
   constructor(emojiElement: HTMLParagraphElement) {
-    this.emojiElement = emojiElement
+    this.emojiElement = emojiElement;
   }
 
-  render() {}
+  abstract render(): void;
 }
 
 class Bee extends Bug<string> {
@@ -14,7 +14,7 @@ class Bee extends Bug<string> {
     this.emoji = "🐝";
   }
 
-  render() {
+  override render() {
     this.emojiElement.innerText = this.emoji;
   }
 }
@@ -25,7 +25,7 @@ class Spider extends Bug<string> {
     this.emoji = "🕷️";
   }
 
-  render() {
+  override render() {
     this.emojiElement.innerText = this.emoji;
   }
 }


jdwilkin4:
Hi @ElectricMolasses ! 

This is off to a good start.

As to your question about what code the camper will start with, they will only start with the HTML and CSS. They will write all the JS/TS code. The only time we really give campers JS code are things like large objects or arrays. So there might be a step or two to have them start building that large array, and then give them the rest in the following step. But in this case, we don't have this situation. They will write all of the JS/TS code themselves. 

So if there are any parts not essential to the learning of this workshop, then yes, it can be removed.

fullstack-cert/typescript-projects/generics-workshop/src/script.ts
```diff
@@ -0,0 +1,109 @@
+// I believe this can be achieved in under 30 steps with the filler code that doesn't have to do with the workshop itself being prefilled. If we want to trim it even more we can look at cutting out the Bug type, or the extra guard on the types of input events. The input guard could be refactored into a function that takes a callback and accepts a Generic for the type of event we want to filter, I think.
+
+// Start off the camper with these three classes
```
naomi-lgbt: We should have them write these classes. It will give them practise adding type annotations and using `extends`.

fullstack-cert/typescript-projects/generics-workshop/src/script.ts
```diff
@@ -0,0 +1,109 @@
...
+
+// Start the camper off with the map
+const bugMap = {
+  "bee": new Bee(),
+  "spider": new Spider(),
+};
+
+// Have the selectors written from the start as well
+const bugEmojiElement = document.querySelector('#bug-emoji') as HTMLParagraphElement;
+const bugNameElement = document.querySelector('#bug-name') as HTMLParagraphElement;
```
naomi-lgbt:
```suggestion
const bugEmojiElement = document.querySelector<HTMLParagraphElement>('#bug-emoji');
const bugNameElement = document.querySelector<HTMLParagraphElement>('#bug-name');
```

It's a workshop teaching generics - we should absolutely have the camper write these and use generics instead of assertions.

fullstack-cert/typescript-projects/generics-workshop/src/script.ts
```diff
@@ -0,0 +1,109 @@
...
+//  modify this class throughout the workshop
+class Renderer<T extends Bug> {
+  currentBug: Bug = bugMap["bee"];
+  name: string = "";
+
+  constructor() {
+    // Have these selectors pre-written for the camper.
+    const bugForm = document.querySelector('#bug-form') as HTMLFormElement;
+    (bugForm.querySelector("#species") as HTMLSelectElement).value = "bee";
+    (bugForm.querySelector("#name") as HTMLInputElement).value = "";
```
naomi-lgbt: Same here. These should be generics not assertions.

fullstack-cert/typescript-projects/generics-workshop/src/script.ts
Rhombert: The workshop is rewritten to be much more minimal, but I've kept the generic base `Bug<T>` class, and the two types of bug now extend it, if in a somewhat contrived way.

fullstack-cert/typescript-projects/generics-workshop/src/script.ts
```diff
@@ -0,0 +1,48 @@
...
+    super(emojiElement);
+    this.emoji = "🕷️";
+  }
+
+  render() {
+    this.emojiElement.innerText = this.emoji;
+  }
+}
+
+function isSelect(element: EventTarget | null): element is HTMLSelectElement {
```
shootermv, line 33: is HTMLSelectElement (may be `as HTMLSelectEleemnt`?)

naomi-lgbt: Nope. This is a function signature, and the `is` keyword allows you to say "Instead of this function returning a boolean, when the function returns `true` then the `element` is of type `HTMLSelectElement`.".

fullstack-cert/typescript-projects/generics-workshop/src/index.html
```diff
@@ -0,0 +1,24 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Generics Lab</title>
```
jdwilkin4: This should probably be changed to better reflect the project like `Bug Form` or something like that.

Rhombert: Renamed to `Bug Emoji Picker`.

fullstack-cert/typescript-projects/generics-workshop/src/script.ts
```diff
@@ -0,0 +1,48 @@
+class Bug<T> {
```
a2937:
```suggestion
abstract class Bug<T> {
```

This class seems to be a good place where abstract classes are introduced. https://www.typescriptlang.org/docs/handbook/2/classes.html#abstract-classes-and-members

fullstack-cert/typescript-projects/generics-workshop/src/script.ts
```diff
@@ -0,0 +1,48 @@
...
+  emoji: T;
```
a2937:
```suggestion
  emoji!: T;
```

This seems like a good way to handle the error message campers could get since `emoji` is never assigned a value. 

https://www.typescriptlang.org/docs/handbook/2/classes.html#--strictpropertyinitialization

a2937:
I am also wondering if you should you should make the render method abstract or simply have it be an inherited method.

fullstack-cert/typescript-projects/generics-workshop/src/script.ts
Rhombert: Applied suggestion

fullstack-cert/typescript-projects/generics-workshop/src/script.ts
Rhombert: I'd like clarification on this error message as I don't believe it should apply here. According to the docs, this error should only appear when it is _not_ definitely assigned in the constructor, but even in the generic base class is is explicitly required.
The example code in the documentation shows assigning it explicitly in the constructor is a fix as well. Does this fix not apply in cases where the assignment comes from an argument as opposed to a static value?

Rhombert:
@a2937 My reasoning for the implementation of the render method only existing in the child classes is that we don't know what type "emoji" will be, and frankly emoji is probably a poor name for the larger idea here. I was writing it with the idea that you could substitute in other types, for example an image, which would break the render method as written on the two string classes.
I then named it emoji to make it easier to understand from a beginner perspective, which in retrospect, creates confusion.
If we do want to move the render functionality to `Bug`, then I think I should also try to find a way to constraint `T` to types that can be dumped into `innerText` and function.

fullstack-cert/typescript-projects/generics-workshop/src/script.ts
Rhombert: Received clarification and tested in VSCode. Definitely seems to be an issue with my own LSP configuration not catching this. Also appears to be an edge case with typescript, since the definition of the error, AND the example of the error provided by the link in VSCode, states that this use case should qualify as being "definitely assigned" in the constructor.

Pushed change that applies the suggested change.

Rhombert: I'm an idiot and can't read, this is working as intended, and the fix is perfectly sane. I just can't read my own code.

a2937 (approved):

a2937:
Actually hold that thought. 

> implementation of the render method only existing in the child classes is that we don't know what type "emoji" will be,

I'm still not sure why it isn't abstract then.

Rhombert:
> Actually hold that thought.
> 
> > implementation of the render method only existing in the child classes is that we don't know what type "emoji" will be,
> 
> I'm still not sure why it isn't abstract then.

That would be my inexperience with typescript classes. Is `render` not considered to be abstract implicitly as a result of the class being marked abstract?

`override` is absolutely something I should be doing, I'll get that update pushed immediately. I imagine you're correct about the `abstract` side as well, I just want to confirm.

Rhombert:
Already found the source and `Bug.render` absolutely should be marked abstract. Thank you!

PR #1034, opened 2025-10-21
chore: update resume builder

jdwilkin4:
This PR cleans up the following

- [x] Updated the header text and preview text to resume builder because that is what the app will be called.
- [x] Updated all instances of cv to resume for consistency
- [x] Removed the footer since it wasn't need and the project was long enough as is
- [x] Refactored the working here functionality because I had issues with it reliably working locally
- [x] Removed the skill bars functionality because skill bars are highly discouraged in the industry. so we shouldn't promote that here. 

Checklist:

### DIFF
--- a/fullstack-cert/typescript-projects/resume-builder/src/App.tsx
+++ b/fullstack-cert/typescript-projects/resume-builder/src/App.tsx
@@ -436,7 +436,7 @@ export const EducationForm: React.FC<EducationFormProps> = ({
                   Start Date *
                 </label>
                 <input
-                  type="month"
+                  type="date"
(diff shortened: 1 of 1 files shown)


a2937 (changes requested):
I just noticed that the LinkedIn information doesn't show up on the preview. 

Also the education section seems to be giving me an invalid date if a date is inputted in the form `Month day, year`. I think it's because the type of the input for the dates is `month` instead of `date`.

a2937 (approved):

a2937:
I was unable to find any issues when interacting with the application.

PR #1055, opened 2025-11-17
chore: add express middleware workshop prototype

### LINKED ISSUE #1047: Create prototype for express middleware workshop
You can review the lessons that come before this workshop on the sheet. But we need a short workshop here so campers can practice what they have learned from the lessons.

Image

zairahira:
Checklist:

Closes #1047


jdwilkin4 (commented):
This seems to simple IMO. I know the issue said short but I feel like campers won't really get enough practice working with middleware. Or really see the need for it. 

I think something more complex will showcase why middleware is used. Also, this opens up the door for conversations around project structure because now we can introduce structuring your code with `routes`, `controllers`,  and `middleware` folders

PR #1066, opened 2025-12-02
Feat: add prototype for `lab-story-fragment-shuffler`

l3onhard:
Checklist:

This PR is part of Naomi’s December 2025 Sprint. Its goal is to develop a prototype for the new Lab `Story Fragment Shuffler`.

Given this lab’s heavy reliance on concepts from the `Loops` and `Arrays` modules, I recommend placing it within the `JavaScript Fundamentals Review` module.

You can find the relevant issue on the freeCodeCamp repository here: [#64112](https://github.com/freeCodeCamp/freeCodeCamp/issues/64112)

### DIFF
--- a/fullstack-cert/js-projects/lab-story-fragment-shuffler/user-stories.md
+++ b/fullstack-cert/js-projects/lab-story-fragment-shuffler/user-stories.md
@@ -7,16 +7,16 @@ You'll work with arrays containing story fragment objects. Each fragment object 
 | Property | Description | Example value |
 |-|-|-|
 | `id` | A positive integer indicating the fragment's position in the story | `3` | 
-| `text` | The actual story content | `"and I use Arch btw"` |
+| `text` | The actual story content | `"and I use Arch btw.\""` |
 
 Here is an example of an array containing story fragments: 
 
 ```
 exampleArray = [
-{ id: 3, text: "\"and I use Arch btw.\"" }, 
+{ id: 3, text: "and I use Arch btw.\"" }, 
 , 
 { id: 1, text: "Naomi said:" }, 
-{ id: 3, text: "\"and I use Arch btw.\"" },
+{ id: 3, text: "and I use Arch btw.\"" },
 ]
 ```
 


fullstack-cert/js-projects/lab-story-fragment-shuffler/user-stories.md
```diff
@@ -0,0 +1,21 @@
+# Summary
+
+In this lab, you will restore a coherent narrative from shuffled, duplicated, or missing story fragments. 
+
+You'll work with a pre-filled array called `shuffledFragments` containing story fragment objects. Each fragment object has the following properties:
+
+- `id`: A positive integer indicating the fragment's position in the story
+- `text`: The actual story content
+
+# User Stories
```
jdwilkin4, line 35: Right now the language used here reads more like a workshop than a lab.

In labs, you want to `you should` language instead

- https://www.freecodecamp.org/learn/javascript-v9/lab-factorial-calculator/build-a-factorial-calculator
- https://www.freecodecamp.org/learn/javascript-v9/lab-repeat-a-string/build-a-string-repeating-function

fullstack-cert/js-projects/lab-story-fragment-shuffler/script.js
```diff
@@ -0,0 +1,111 @@
...
+];
+
+const compactFragments = (fragments) => {
+  // guards against sparse arrays by ignoring empty elements and logging warnings
+  let compacted = [];
+
+  for (let i = 0; i < fragments.length; i++) {
+    if (fragments[i] === undefined) {
+      console.log(
+        `WARNING: Undefined element in the array at index ${i}.`,
```
jdwilkin4: it will be easier from  a testing perspective to return the warning string instead of having this console.log statement. 

then when you call the function you can log that function call.

fullstack-cert/js-projects/lab-story-fragment-shuffler/user-stories.md
```diff
@@ -0,0 +1,21 @@
+# Summary
+
+In this lab, you will restore a coherent narrative from shuffled, duplicated, or missing story fragments.
```
jdwilkin4: I think there needs to be more context here.
After reading through the intro and user stories, I don't think campers will fully understand what they are building.
It would be better to have an example story fragment and then the assembled story to show so they get the full picture better.

jdwilkin4:
Also, please update your post to say where this is supposed to go in the curriculum.
JS fundamentals module? Objects module? somewhere else?

fullstack-cert/js-projects/lab-story-fragment-shuffler/script.js
l3onhard: I was thinking we could try the following approach using the `spyOn` helper. Would that work for you?

**Here is the adapted user story:**

1. You should create a function named `compactFragments` that takes an array of fragments and returns a new array with all undefined elements removed. If the function removes any undefined elements, it should log a message to the console. The message should start with the prefix `[COMPACTED]`.

**Here is the adapted function from the solution:**

```ja
const compactFragments = (fragments) => {
  let compacted = [];
  let numberOfUndefined = 0;

  for (const fragment of fragments) {
    if (fragment === undefined) {
      numberOfUndefined++;
    } else {
      compacted.push(fragment);
    }
  }

  if (numberOfUndefined > 0) {
    console.log(`[COMPACTED] Removed ${numberOfUndefined} undefined fragment(s).`);
  }

  return compacted;
};
```

**Here is the hint/test using `spyOn`**

When the function `compactFragments` is called with the `shuffledFragments` array. I should log a message starting with `[COMPACTED]` to the console.

```js
const spy = __helpers.spyOn(console, 'log');
compactFragments(shuffledFragments);
assert.match(spy.calls[0], /^\[COMPACTED\]/);
spy.restore();
```

fullstack-cert/js-projects/lab-story-fragment-shuffler/user-stories.md
l3onhard: I added an example to the intro: [prototype: add more context to the intro in story-fragment-shuffler lab](https://github.com/freeCodeCamp/CurriculumExpansion/pull/1066/commits/d391821bc4790696ad7f6a22ed86efdc58884d8a)

fullstack-cert/js-projects/lab-story-fragment-shuffler/user-stories.md
l3onhard: Done: [prototype: adapt user story language in story-fragment-shuffler lab](https://github.com/freeCodeCamp/CurriculumExpansion/pull/1066/commits/8fe9e354b0c6c020256dbcdb879d3f2f15c947a2)

fullstack-cert/js-projects/lab-story-fragment-shuffler/script.js
jdwilkin4: yeah, that should work

PR #1078, opened 2025-12-15
feat: add a prototype for the Guild Loot Tracker workshop

Sebastian-Wlo:
Checklist:

Relates to: https://github.com/freeCodeCamp/freeCodeCamp/issues/64119

- link to the related issue: https://github.com/freeCodeCamp/freeCodeCamp/issues/64119 ,
- currently, the prototype uses `for()` loops and `Array.prototype.sort()` method,
- I couldn't figure out how to use "Object.values()" in the functions (maybe adding a function returning the sums of ingredient macros to the requirements would be a good place to implement it?)


jdwilkin4 (commented):
Please make to include the user stories so those can be review too just in case updates need to be made there. 

Also, please let us know where this falls in the curriuclum.

loops module?
review modules?
higher order functions?

 we will need that information when conducting reviews 👍🏾

Sebastian-Wlo:
Just to make sure: should the user stories be a copy of those mentioned in the related issue, or should I come up with some new ones?

As for where this lab should be placed in the curriculum - originally, it was supposed to be in the Objects module. But I think `listTopIngredients()`, which was supposed to return sorted output requires the use of a loop. `sort()` would be ideal, but it should be doable with just a `for()` loop (I'd definitely need to make some changes to make it work).

So, Loops would probably be the right place for it.

Thank you, I'll add the user stories as soon as I can.

majestic-owl448:
you can use those from the issue as base, and update as needed based on changes that have been made

Sebastian-Wlo:
Sorry! I tried to rename the branch without thinking what it would do to the PR.

`user-stories.md` has been added

majestic-owl448:
wouldn't this project have a similar issue to the nutrient tracker that we are replacing because potentially triggering?

Sebastian-Wlo:
> wouldn't this project have a similar issue to the nutrient tracker that we are replacing because potentially triggering?

Possibly, in the current form at least - someone actually pointed it out after I chose the issue.
It's possible that I misinterpreted the response, but the way I understood it was that this _could_ still be useful, but maybe after changing its subject.

I'm not sure if it means I should come up with a different theme (I'm not the most creative person out there, so it might take some time), or leave the prototype as if for now and wait for a review.

majestic-owl448:
there are many other things that could be tracked in a pretty similar way, so the prototype is still good, there will be need to change the name of some things before it's ready tho

Sebastian-Wlo:
The overall "theme" has been changed to a new one (thank you once more @naomi-lgbt, if it wasn't for you, I would have been stuck at coming up with something for some time).

It will definitely need changes, since `getMemberTotals()` is the only function that returns a string while the other return objects (incidentally, it is unfortunately the only point where I managed to find a use for `Object.values()` in). I hope that at least one approach is good, and can be applied to the rest of the functions.

Should I request another review for this?

fullstack-cert/js-projects/lab-guild-loot-tracker/user-stories.md
```diff
@@ -0,0 +1,13 @@
...
+
+**User Stories:**
+
+1. You should accept meal entries shaped like `{ member, gold, silver, reputation, experience }`.
+2. You should implement `addLootEntry(object, entry)` that clones and updates cumulative resources per member.
+3. You should implement `getMemberTotals(object, member)` with guard clauses for unknown members.
+4. You should implement `listTopMembers(object, key, limit)` that sorts descending by a specific resource type (e.g., finding the member with the most `gold`).
+5. You should implement `cloneGuildData(object)` to protect against accidental mutations.
+
+Note: Use `Object.keys`, `Object.values`, and `Object.entries`. Include tests for aggregation accuracy and cloning behavior.
```
jdwilkin4: Is this just a note for yourself? Or part of the user stories?
Also, please double check in the curriculum if `Object.keys`, `Object.values`, and `Object.entries` has been taught by this point. My gut reaction is no. If that is the case, then you will need to update the approach for this lab.

fullstack-cert/js-projects/lab-guild-loot-tracker/user-stories.md
```diff
@@ -0,0 +1,13 @@
+In this lab, you will aggregate resource totals per guild member inside nested objects.
+
+**Objective:** Fulfill the user stories below and get all the tests to pass to complete the lab.
+
+**User Stories:**
+
+1. You should accept meal entries shaped like `{ member, gold, silver, reputation, experience }`.
```
jdwilkin4: This should be updated too. I think there needs to be more information for campers on what this means because I can see a lot of campers being confused on what to do here.

fullstack-cert/js-projects/lab-guild-loot-tracker/user-stories.md
```diff
@@ -0,0 +1,13 @@
...
+2. You should implement `addLootEntry(object, entry)` that clones and updates cumulative resources per member.
+3. You should implement `getMemberTotals(object, member)` with guard clauses for unknown members.
+4. You should implement `listTopMembers(object, key, limit)` that sorts descending by a specific resource type (e.g., finding the member with the most `gold`).
+5. You should implement `cloneGuildData(object)` to protect against accidental mutations.
```
jdwilkin4: this will need more information to in regards to what protecting against accidental mutations means for beginners.

fullstack-cert/js-projects/lab-guild-loot-tracker/user-stories.md
```diff
@@ -0,0 +1,13 @@
+In this lab, you will aggregate resource totals per guild member inside nested objects.
```
jdwilkin4: This introduction needs more context on what to build. otherwise campers won't really understand what guild members are and what the end goal for this project is.

jdwilkin4 (commented):
I left some thoughts on the user stories.
But also noticed that none of these functions are being used anywhere. 

From an educational standpoint, we don't want to send the message to beginners to create functions in their code and not use it. All of these functions should be used in some way so campers can connect the dots and see what is actually happening in the code. Even if it is just logging these function call results to the console.

jdwilkin4:
Also, please update your PR title since the project was changed 👍🏾

fullstack-cert/js-projects/lab-guild-loot-tracker/user-stories.md
Sebastian-Wlo: I copied the current user stories from the original issue - the contribution guide mentions the additions of notes/tips, so I included the note from them as well. Since I've only contributed to some quiz questions before, trying to match the initial requirements as close as I could seemed safer than rewriting them (i hope that makes at least some sense). 

Thank you, I'll go over the other labs in this part of the curriculum and try to make the user stories more descriptive and uniform with those.

fullstack-cert/js-projects/lab-guild-loot-tracker/user-stories.md
Sebastian-Wlo: Oh, I completely missed the `meal` reference on that line...
I tried to add function explanations to the lab's summary to make room for more specific requirements in the user stories (which I haven't really touched yet).
But now I think about it, the correct place for the requirements is probably in the descriptions of the checks, right? So, should I try to turn those explanations into user stories?

jdwilkin4 (commented):
I can't find where Object.keys, Object.values, and Object.entries is taught in the JS curriculum.

A new lesson can be added to the JS fundamentals module for that. 
But until that can happen, we won't be able to add this lab to the curriculum because we can't ask campers to build projects on topics they haven't learned yet. And this lab seems to be solely dedicated to practicing those methods mentioned earlier. 

So you will need to reach out to Naomi on next steps. Either we are going to wait until the new lesson has been added so this lab can be added. Or not add this new lab at all.

Sebastian-Wlo:
> But until that can happen, we won't be able to add this lab to the curriculum because we can't ask campers to build projects on topics they haven't learned yet. And this lab seems to be solely dedicated to practicing those methods mentioned earlier.

That's fair - I've been trying to make the prototype work without these methods, and while it could be done, it felt like it wasn't introducing anything that wasn't already shown in the labs for the loops. It also felt more "awkward" than it probably should, if that makes sense, but I wasn't sure if that was because the lab's description didn't really fit that section, or if it was just me doing a bad job of it.

> So you will need to reach out to Naomi on next steps. Either we are going to wait until the new lesson has been added so this lab can be added. Or not add this new lab at all.

I'll do that, thank you!

jdwilkin4:
@Sebastian-Wlo 

What's the status for this lab?

Sebastian-Wlo:
@jdwilkin4
Sorry, forgot to forward it here - if I understood correctly, it might get turned into a workshop for using `.entries()`, `.keys()` and `.values()` at some point.

I haven't worked on it in a while, but I can try modifying the prototype to use these specific methods instead of loops so it would be ready to be divided into steps.

jdwilkin4:
I would confirm with Naomi to see what next steps should be here 👍🏾

Sebastian-Wlo:
I asked her a while ago, this is what she told me:
> The only place we are using a loop is in that "listTopMembers" function.
> What if we:
> 1. Change that to listMembers and just print the Object.entries or something simple.
> 2. That means we could put this in the objects section instead of the loops section, I think?
>3. Which then means we could turn it into a workshop and teach those concepts there!

... which, had I read the response *properly* back then, I could have done a long time ago. I'll make the changes soon.

Sebastian-Wlo:
Actually, as far as I understand, turning this lab into a workshop would probably require going through a completely different process - I'll ask Naomi how to do it properly and if I should create, or wait for, an issue about the lab.
Sorry, I've been a little distracted lately.

majestic-owl448:
at the prototype stage, the main difference is that a workshop does not need user stories, and that the protytpe is the final solution of the project, not just a possible solution

jdwilkin4:
@Sebastian-Wlo 

Can you post any updates you have regarding this project?
Thanks 👍🏾

jdwilkin4:
It looks like a quick glance there were recent conversations on discord regarding this. But those conversations can easily be burried.
So it is best to post an update here so we can track of the progress all in one place 👍🏾

Sebastian-Wlo:
@jdwilkin4, Sorry for the lack of any activity!
I know that this will get turned into a workshop - so, I need to remove the remaining `for()` loop, make sure that the prototype is cleanly written and get rid of the user stories before planning out how to break the code into steps.
I'll do my best to update the prototype over the weekend, or in the worst case early next week.

Sebastian-Wlo:
First of all, I'm really sorry it took me so long to add any updates.

I'm still not happy with the `addLootEntry()` and `cloneGuildData()` functions. I understand what purpose creating a shallow copy of the object is _supposed_ to have, but it currently doesn't really add anything.
Another thing that doesn't seem right is the inconsistency between returning values and using `console.log()`s in the functions. I'm not sure which would work better - it seems like returning strings and logging them to the console would be the best solution, but I'm not sure if this makes the most sense here.

I hope to keep working on this, but I'd really appreciate any pointers on how to turn this into a actually useful workshop

jdwilkin4:
@Sebastian-Wlo 

it would probably be best to open up a thread on discord under naomi's sprints channel.
That way we can hash this out get more clarity on the core objectives for this lab.

Sebastian-Wlo:
Terribly sorry it took so long! I've unfortunately been really busy.

Sorry the prototype is not up to an acceptable level. Aside from from all of the (too many) `TODO`s in the current version, are there any glaring mistakes that need to be addressed?

During the conversation in the Naomi's sprints channel it turned out that it's not clear the names I've used for the guild members are actually names (which I should have seen coming, to be honest).
I was thinking about using the names of the fCC staff, like it's been done in some of the previous workshops, but it didn't feel comfortable with assigning people reputation and experience arbitrarily. Would it be okay if I did that? I don't want anyone to be offended about the "stats" I'm assigning here

fullstack-cert/js-projects/guild-loot-tracker/script.js
```diff
@@ -0,0 +1,157 @@
+// This will be a workshop for the Loops section, probably after the "Build a Profile Lookup" lab.
+ 
+// Creating the "guild" object could be 2 or 3 Steps (creating "guild" object itself and adding the first "guild member" object as the first property - rest of the "guild members" can be filled out somewhere between the next steps to avoid too much repetition).
+let guild = {
```
jdwilkin4, line 5: these names should work

```js
let guild = {
  ethan: {
    gold: 31,
    silver: 48,
    reputation: 9,
    experience: 198,
  },
  elara: {
    gold: 78,
    silver: 64,
    reputation: 12,
    experience: 111,
  },
  brandon: {
    gold: 41,
    silver: 7,
    reputation: 7,
    experience: 70,
  },
  dylan: {
    gold: 81,
    silver: 2,
    reputation: 20,
    experience: 220,
  },
  lucas: {
    gold: 34,
    silver: 28,
    reputation: 10,
    experience: 179,
  },
  natalie: {
    gold: 36,
    silver: 81,
    reputation: 12,
    experience: 82,
  },
};
```

jdwilkin4:
> I was thinking about using the names of the fCC staff, like it's been done in some of the previous workshops, but it didn't feel comfortable with assigning people reputation and experience arbitrarily. Would it be okay if I did that? I don't want anyone to be offended about the "stats" I'm assigning here

I left a comment regarding the names. But in general, some of the staff would probably be fine having their names in the curriculum while others might not. So using fake yet more realistic names should be fine here.

Sebastian-Wlo:
Thank you @jdwilkin4 , that list solves my problem with names!
I still need to clean up the comments and note how to divide the code into steps (I'll try to do that within the next day or two), but **most** of the current version is hopefully ready for the review.

One thing I was thinking about is getting rid of the `listMemberNames()` function and introducing both `Object.keys()` and `Object.values()` in the `listMembers()`.
I'll stick to making any suggested changes (instead of rewriting functions because I randomly came up with a better solution), but I'd like to know if removing `listMemberNames()` is a good idea, or if introducing these two methods in separate functions would be a cleaner way to present them.

jdwilkin4:
> but I'd like to know if removing listMemberNames() is a good idea, or if introducing these two methods in separate functions would be a cleaner way to present them.

We can try the separate functions and then through the review process if we decide it doesn't work as well, we go back to the `listMemberNames`.  👍🏾

Sebastian-Wlo:
Once again, sorry it took so long - I was too optimistic about how much time I'd have this week... On the upside, I _think_ all of the changes I wanted to make are done, and hopefully the steps breakdown is reasonably clear.
I don't expect the prototype to pass the review on the first go, but hopefully some of it will be usable:)

jdwilkin4 (commented):
After reviewing this and seeing this will be a workshop after the profile lookup lab, I would suggest simplifying this and keeping the core learning concepts to working with Object.keys(), Object.values(), and Object.entries()

Here is my proposed update

```js
const guild = {
  ethan: {
    gold: 31,
    reputation: 9,
  },
  elara: {
    gold: 78,
    reputation: 12,
  },
  brandon: {
    gold: 41,
    reputation: 7,
  },
  dylan: {
    gold: 81,
    reputation: 20,
  },
};

function listMembers(guildObject) {
  console.log("Guild Members:");

  const members = Object.keys(guildObject);

  for (const member of members) {
    console.log(member);
  }
}

listMembers(guild);

console.log("------------");

function countMembers(guildObject) {
  return Object.keys(guildObject).length;
}

console.log(`Guild Members: ${countMembers(guild)}`);

console.log("------------");

function listGold(guildObject) {
  console.log("Gold Totals:");

  for (const [name, stats] of Object.entries(guildObject)) {
    console.log(`${name}: ${stats.gold} gold`);
  }
}

listGold(guild);

console.log("------------");

function getTotalGold(guildObject) {
  let totalGold = 0;

  for (const member of Object.values(guildObject)) {
    totalGold += member.gold;
  }

  return totalGold;
}

console.log(`Total Gold: ${getTotalGold(guild)}`);

console.log("------------");

function getRichestMember(guildObject) {
  let richestName = "";
  let highestGold = 0;

  for (const [name, stats] of Object.entries(guildObject)) {
    if (stats.gold > highestGold) {
      highestGold = stats.gold;
      richestName = name;
    }
  }

  return richestName;
}

console.log(`Richest Member: ${getRichestMember(guild)}`);

console.log("------------");

function listVeterans(guildObject) {
  console.log("Veteran Members:");

  for (const [name, stats] of Object.entries(guildObject)) {
    if (stats.reputation >= 10) {
      console.log(name);
    }
  }
}

listVeterans(guild);

console.log("------------");

function awardGold(guildObject, amount) {
  for (const member of Object.values(guildObject)) {
    member.gold += amount;
  }
}

awardGold(guild, 10);

console.log("After Awarding 10 Gold:");

listGold(guild);

console.log("------------");

function guildReport(guildObject) {
  console.log("Guild Report");
  console.log("------------");
  console.log(`Members: ${countMembers(guildObject)}`);
  console.log(`Total Gold: ${getTotalGold(guildObject)}`);
  console.log(`Richest Member: ${getRichestMember(guildObject)}`);
}

guildReport(guild);

```

Sebastian-Wlo:
That's quite a departure from the current version, but it makes sense considering the whole idea was moved to a different section, as well as from a lab to a workshop:)
It's a shame that object cloning didn't make the cut,  but it makes sense to keep the scope of the workshop more focused on a specific subject. 

I like how much cleaner it looks! But there are 2 things I'd like to ask about before rewriting it:
1. Should `reputation` be left without any function modifying the values? It's fine, it's an exercise and not an actual management app after all (and it keeps this workshop less repetitive), but it does look a little odd to have a value and no way to change it,
2. Would it be ok to separate the "headers" of the function responses (so, "Guild Members:", "Gold Totals:" etc) from the rest of the output with `------------` and functions with just `console.log("")`? Either that, or removing the lines altogether - they look nice, but it's easy to get the number of characters wrong, so maybe it would be better to avoid adding them in multiple places

jdwilkin4:
> Should reputation be left without any function modifying the values? It's fine, it's an exercise and not an actual management app after all (and it keeps this workshop less repetitive), but it does look a little odd to have a value and no way to change it,

I think it is fine to leave reputation. 

> Either that, or removing the lines altogether - they look nice, but it's easy to get the number of characters wrong, so maybe it would be better to avoid adding them in multiple places

If you want to remove them, that is fine

majestic-owl448:
@Sebastian-Wlo do you have any update on this prototype?

Sebastian-Wlo:
@majestic-owl448 yes, sorry, I've been very busy recently. I have started re-writing it as per Jessica's proposed update (with added step breakdown) and will try to push the changes in the next 48 hours.

majestic-owl448:
it looks like this is intended to go in the objects module, right?

if that so, keep in mind loops are introduced only later, if it goes in loops then everything ok

image

Sebastian-Wlo:
From what I understand it should end up in the "Loops" section, somewhere after the "Build a Profile Lookup" Lab - which, on one hand does have a lot of steps already, but on the other has a lot of Labs back to back. I hope that's ok

majestic-owl448:
yes, it's fine, it also has practice of something we don't have much of

Sebastian-Wlo:
I started writing instructions for each step, and the current version can get a bit to packed at times. Not a problem with the code itself, but the step breakdown in the comments could be better.
The final version shouldn't end up with more than 35 steps, but it's still more than currently indicated.

I'm not sure if I should fix the comments now, since I'd imagine that when I get to actually breaking the prototype into steps there will be some other minor changes to the pacing anyway.

jdwilkin4:
> I'm not sure if I should fix the comments now, since I'd imagine that when I get to actually breaking the prototype into steps there will be some other minor changes to the pacing anyway.

I wouldn't worry to much about the steps and descriptions at this stage. 
The goal of this stage is to have a working prototype that the maintainers can review that works in the current place in the curriculum.

Once the prototype is finalized, then you can worry about steps and tests. Also, in the next review stage in the main repo, maintainers will probably have suggestions for breaking things down further if needed.

jdwilkin4:
@Sebastian-Wlo 

If this is ready for review, then we can take another look at it.
If you are still working on this, then continue to push up commits and then let use know when this is ready for review.

Sebastian-Wlo:
@jdwilkin4 It is ready for another review - I can't think of anything that would improve the code at this point (it is almost exactly your simplified version of the workshop line per line - also thank you again for writing it! It's definitely better suited to be turned into a Workshop than my previous attempts).
I actually forgot GitHub has a 'Re-request a review' option.

imsomilg:
I think we should have `console.log("------------");` instead of `console.log("")`. Maybe we can create a function and keep reusing it so that users don't have to count no. of dashes every time?

fullstack-cert/js-projects/guild-loot-tracker/script.js
```diff
@@ -0,0 +1,172 @@
...
+  for (const member of Object.values(guildObject)) {
+    totalGold += member.gold;
+  }
+
+  return totalGold;
+}
+
+console.log(`Total gold: ${getTotalGold(guild)}`);
+
+console.log("");
```
majestic-owl448:
```suggestion
console.log(`Total gold: ${getTotalGold(guild)}\n`);
```

fullstack-cert/js-projects/guild-loot-tracker/script.js
```diff
@@ -0,0 +1,172 @@
...
+      richestName = name;
+    }
+  }
+
+  return richestName;
+}
+
+console.log(`Richest Member: ${getRichestMember(guild)}`);
+
+console.log("");
```
majestic-owl448:
```suggestion
console.log(`Richest Member: ${getRichestMember(guild)}\n`);
```

majestic-owl448 (commented):
I don't really like having `console.log("");`, it does the same as `console.log()`, and can be replaced at least in some places with a `\n` in the previous step

(it would also make for a boring step, to add an empty console.log, tbh)

(yeah, it's a nitpick, I would say it's ready)

Sebastian-Wlo:
Both are good solutions - thank you for your suggestions! I think adding a function for logging a "separator line" would work slightly better, since it will not require any workarounds for functions used in `guildReport()` to look uniform.

Honestly, I went with logging an empty string instead of adding `\n` mainly because asking for a newline character could look a bit confusing in the instructions. Adding it at the end of `"Guild Report"` shouldn't be as much of a problem though, considering it can be given in the step as a whole string.
It would also allow to differentiate the line below the "header" in the `guildReport` and separators between the earlier functions.

jdwilkin4 (approved):
I think these changes look good.

PR #1081, opened 2025-12-17
feat(workshop): add prototype for Weather Trend Analyzer

aliyasyeddd:
Checklist:

  - [X] My pull request has a [descriptive title](https://contribute.freecodecamp.org/#/how-to-open-a-pull-request?id=prepare-a-good-pr-title) (not a vague title like `Update index.md`)

Relates to https://github.com/freeCodeCamp/freeCodeCamp/issues/64124
This is part of Naomi's sprints for the JavaScript/Full-Stack certificate.

### DIFF
--- a/fullstack-cert/js-projects/weather-trend-analyzer/script.js
+++ b/fullstack-cert/js-projects/weather-trend-analyzer/script.js
@@ -94,7 +94,7 @@ function findMissingTemps(readings) {
     return [];
   }
 
-  const HOUR = 60 * 60 * 1000;
+  const hour = 60 * 60 * 1000;
   const missing = [];
 
   let current = readings[0].time;
@@ -110,23 +110,27 @@ function findMissingTemps(readings) {
       }
     }
 
-    if (found === false) {
+    if (!found) {
       missing.push(current);
     }
 
-    current += HOUR;
+    current += hour;
   }
 
   return missing;
 }
 
 
+// NOTE TO INSTRUCTOR:
+// This step uses `console.warn`.
+// Be sure to introduce and explain `console.warn`
+// earlier in the workshop before campers reach this step.
 function summarizeTrends(readings) {
-  const HEAT_THRESHOLD = 38;
-  const COLD_THRESHOLD = 20;
+  const heatThreshold = 38;
+  const coldThreshold = 20;
 
-  const heatWaves = detectHeatWave(readings, HEAT_THRESHOLD);
-  const coldSnaps = detectColdSnap(readings, COLD_THRESHOLD);
+  const heatWaves = detectHeatWave(readings, heatThreshold);
+  const coldSnaps = detectColdSnap(readings, coldThreshold);
   const missing = findMissingTemps(readings);
 
   if (heatWaves.length > 0) {
@@ -149,3 +153,5 @@ function summarizeTrends(readings) {
   };
 }
 
+const summary = summarizeTrends(temps);
+console.log("Temperature summary:", summary);


majestic-owl448:
please link to the related issue, so reviewers know the contenxt

package-lock.json
majestic-owl448, line 1: please remove the `package-lock.json` file from the PR

jdwilkin4 (commented):
Where does this go in the curriculum?
the loops module?
review JS module?
somewhere else?

we will need to know that information to be able to review accordingly 👍🏾

majestic-owl448:
from Naomi's plan, it's loops
image

majestic-owl448:
you don't need to do merge commits this often, be aware that each one will send a notification to maintainers

aliyasyeddd:
@majestic-owl448 Thanks for the heads up! I’ll avoid frequent merge commits and will keep that in mind going forward.

fullstack-cert/js-projects/weather-trend-analyzer/script.js
```diff
@@ -0,0 +1,151 @@
+
+
+const temps = [
```
jdwilkin4, line 3: since this is a workshop, is the plan to have them build out this whole array of objects? or just part of it, and then provide the rest?

fullstack-cert/js-projects/weather-trend-analyzer/script.js
```diff
@@ -0,0 +1,151 @@
...
+    }
+
+    current += HOUR;
+  }
+
+  return missing;
+}
+
+
+function summarizeTrends(readings) {
```
jdwilkin4, line 128: this function doesn't seem to be used anywhere.
it needs to be called so campers can see what is going on in this workshop.

fullstack-cert/js-projects/weather-trend-analyzer/script.js
```diff
@@ -0,0 +1,151 @@
...
+  const HEAT_THRESHOLD = 38;
+  const COLD_THRESHOLD = 20;
+
+  const heatWaves = detectHeatWave(readings, HEAT_THRESHOLD);
+  const coldSnaps = detectColdSnap(readings, COLD_THRESHOLD);
+  const missing = findMissingTemps(readings);
+
+  if (heatWaves.length > 0) {
+    console.warn("Heat wave detected");
```
jdwilkin4, line 137: I don't believe console.warn has been taught yet.
It is fine to teach it here, but you will need to leave a note for yourself to make sure you have an introduction for this in one of the steps.

fullstack-cert/js-projects/weather-trend-analyzer/script.js
```diff
@@ -0,0 +1,151 @@
...
+    let found = false;
+
+    for (let i = 0; i < readings.length; i++) {
+      if (readings[i].time === current) {
+        found = true;
+        break;
+      }
+    }
+
+    if (found === false) {
```
jdwilkin4: I think `if (!found)` would be better here.

fullstack-cert/js-projects/weather-trend-analyzer/script.js
```diff
@@ -0,0 +1,151 @@
...
+  return result;
+}
+
+
+function findMissingTemps(readings) {
+  if (readings.length < 2) {
+    return [];
+  }
+
+  const HOUR = 60 * 60 * 1000;
```
jdwilkin4: We should change all of these uppercase variable names to lowercase. Reason being is because it will be confusing to a beginner if you use uppercase in some cases and lowercase for others when using const. 

Even though this is a convention used in some projects to signal a constant, we should use consistent naming throughout the curriculum.

fullstack-cert/js-projects/weather-trend-analyzer/script.js
```diff
@@ -0,0 +1,151 @@
...
+    current += HOUR;
+  }
+
+  return missing;
+}
+
+
+function summarizeTrends(readings) {
+  const HEAT_THRESHOLD = 38;
+  const COLD_THRESHOLD = 20;
```
jdwilkin4: see earlier note uppercase variable names.

fullstack-cert/js-projects/weather-trend-analyzer/script.js
aliyasyeddd: Agreed — consistency is more important here than signaling constants. Using lowercase variable names for all const declarations will be clearer for beginners and keep naming consistent across the curriculum. I’ll update these to lowercase throughout.

fullstack-cert/js-projects/weather-trend-analyzer/script.js
aliyasyeddd: Good suggestion — I agree. if (!found) is more idiomatic and easier to read here. I’ll update it to use that form.

fullstack-cert/js-projects/weather-trend-analyzer/script.js
aliyasyeddd: Good call — you’re right that console.warn hasn’t been introduced yet.  I’ll leave a note for myself to make sure it’s explicitly introduced and explained in an earlier step of the workshop so campers aren’t surprised by it.

fullstack-cert/js-projects/weather-trend-analyzer/script.js
aliyasyeddd: Good catch — you’re right. To make the workshop clearer, we should explicitly call this function so campers can see how everything ties together. Adding a call at the bottom of the file (for example, invoking it with the provided temps array and logging the result) will give learners visible output and a clear entry point for the workshop.

fullstack-cert/js-projects/weather-trend-analyzer/script.js
aliyasyeddd: Since this is a workshop, the intent is not for campers to build out the entire array of objects themselves. Per the user story (“You should ingest an array of daily temperature readings”), the array will be provided, and the focus is on consuming and processing that data rather than constructing it. At most, they might add or inspect a small example entry to understand the data shape, but the bulk of the array should be supplied for them.

jdwilkin4: The current plan is to provide it for them.
Keep in mind though that in the next review stage when more curriculum team members get involved they might have differing opinions on this and might want to have campers build out the full thing.

My suggestion would be to write the steps with the array already present in step 1, but be prepared for possible updates in the review process 👍🏾

jdwilkin4 (approved):
This looks good to go in.

Here is what the next part of the process will look like.

Since this is a workshop, please review the docs first 
https://contribute.freecodecamp.org/how-to-work-on-workshops/

You will now work on creating a PR in the main repo with steps and tests.

Start with just outlining the steps and create a draft PR. Then add hints and tests. Make sure to test everything locally before marking it ready for review. You should be able to go through each step locally and making sure everything passes. 

You should also run the test suite 
https://contribute.freecodecamp.org/how-to-work-on-coding-challenges/#testing-challenges

Major changes shouldn't be needed for the final solution. But small changes might be suggested in these code reviews since more people will be involved in this stage. So be prepared for that.

fullstack-cert/js-projects/weather-trend-analyzer/script.js
aliyasyeddd: That makes sense 👍
I’ll proceed with the array already provided in step 1 for now, in line with the current plan and the user story. I’ll also keep the steps flexible and be prepared to adjust if the curriculum team prefers having campers build out the full array during the next review stage.

PR #1082, opened 2025-12-18
feat(curriculum): create prototype for Python functions workshop

jacobsandersen:
Checklist:

This PR is part of Naomi's Python Basics sprint, and is creating the Python functions workshop prototype. I think this is pedagogically sound, but please review and let me know if you have any improvement suggestions. 

I am not sure if by this point users have been introduced to the tuple datatype; if not, it might be better if I change one of the functions to not use them.

Relates to https://github.com/freeCodeCamp/freeCodeCamp/issues/64627

### DIFF
--- a/fullstack-cert/python-projects/functions-workshop/main.py
+++ b/fullstack-cert/python-projects/functions-workshop/main.py
@@ -12,11 +12,11 @@ def take_inventory():
     # This should clarify that total_items is accessible only inside take_inventory
     total_items = eggs + flour + sugar
 
-    # Step 4, review printing and string concatenation referencing local and global scoped variables
-    print("The kitchen has", total_items, "total items:")
-    print("- ", eggs, "eggs")
-    print("- ", flour, "flour")
-    print("- ", sugar, "sugar")
+    # Step 4, review printing and f-strings referencing local and global scoped variables
+    print(f'The kitchen has {total_items} total items:')
+    print(f'- {eggs} eggs')
+    print(f'- {flour} flour')
+    print(f'- {sugar} sugar')
 
 # Step 5, call a function and immediately see results
 take_inventory()
@@ -27,7 +27,7 @@ take_inventory()
 def use_eggs(eggs, num_eggs):
     # Step 7, review conditionals and local/global scopes
     if num_eggs > eggs:
-        print("The kitchen does not have enough eggs.")
+        print('The kitchen does not have enough eggs.')
 
         # Step 8, explain that returning in a function ends the function call and returns control
         # to the original call-site (below)
@@ -35,10 +35,10 @@ def use_eggs(eggs, num_eggs):
 
     # Step 9, compute the new value of eggs and return it
     # Opportunity to stop and note:
-    # - But why do we need to accept "eggs" as a parameter? Why can't we just modify it?
+    # - But why do we need to accept `eggs` as a parameter? Why can't we just modify it?
     # - Q: Should we teach that `global` exists, but that it is generally considered bad practice?
-    else:
-        return eggs - num_eggs
+    print(f'{num_eggs} egg(s) used out of {eggs} available.')
+    return eggs - num_eggs
 
 # Step 10, introduce calling a function and giving it arguments
 use_eggs(eggs, 1)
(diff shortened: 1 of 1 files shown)


majestic-owl448:
tuples are not introduced, you can refer this to know https://github.com/freeCodeCamp/freeCodeCamp/issues/64537#issuecomment-3647068036

you can also refer to the certification itself, we are in Python Basics, tuples are taught after. I do not see tuples in the prototype tho

jacobsandersen:
> tuples are not introduced, you can refer this to know https://github.com/freeCodeCamp/freeCodeCamp/issues/64537#issuecomment-3647068036
> 
> 
> 
> you can also refer to the certification itself, we are in Python Basics, tuples are taught after. I do not see tuples in the prototype tho

Ah okay, thanks. In the `use_eggs` function, I return tuples. I'll change this later.

fullstack-cert/python-projects/functions-workshop/main.py
```diff
@@ -0,0 +1,74 @@
...
+        # Step 8, explain that returning in a function ends the function call and returns control
+        # to the original call-site (below)
+        return eggs
+
+    # Step 9, compute the new value of eggs and return it
+    # Opportunity to stop and note:
+    # - But why do we need to accept "eggs" as a parameter? Why can't we just modify it?
+    # - Q: Should we teach that `global` exists, but that it is generally considered bad practice?
+    else:
+        return eggs - num_eggs
```
Dario-DC: In a following step the `else` can be removed to let campers understand that the function either runs the `if` and returns or runs the code below.

Also, it could print something when eggs are used.

So the function becomes:
```py
def use_eggs(eggs, num_eggs):
    if num_eggs > eggs:
        print('The kitchen does not have enough eggs.')
        return eggs
    
    print(f'{num_eggs} egg(s) used out of {eggs} available.')
    return eggs - num_eggs
```

Dario-DC (commented):
Hey @jacobsandersen 👋 I left a comment, otherwise it looks good to me. Just remember that we use single quotes for Python. It's not important to update them here, but in case you take care of adding the project to main, that is something to remember.

fullstack-cert/python-projects/functions-workshop/main.py
```diff
@@ -0,0 +1,74 @@
...
+sugar = 3
+
+# Step 2, write a basic function to determine kitchen inventory
+# Demonstrates local and global scope
+def take_inventory():
+    # Step 3, define a local scoped variable
+    # This should clarify that total_items is accessible only inside take_inventory
+    total_items = eggs + flour + sugar
+
+    # Step 4, review printing and string concatenation referencing local and global scoped variables
```
majestic-owl448: here strings are not being concatenated, maybe the first `print` should be so

majestic-owl448 (approved):
I would say it's good too, just left a comment for consideration

fullstack-cert/python-projects/functions-workshop/main.py
jacobsandersen: Agreed. I was deliberating about whether to use an else or not. I will remove it.

fullstack-cert/python-projects/functions-workshop/main.py
jacobsandersen: Thanks for calling this out. I've switched instead to f-strings, since I believe students should be aware of them by this point in the course (according to Dario's earlier comment).

jacobsandersen:
Thanks for the reviews, comments, and suggestions all. I've made some further changes to align to fCC's code style, make the steps a bit clearer for learners, and fix an inconsistency between the comment and code in an early step.

majestic-owl448:
once this is merged, you can start on adding the workshop to the freeCodeCamp/freeCodeCamp repo, start with the structure with the `create-new-project` script and open a draft PR, after that start creating steps and write their description and seed code, and the final solution in the last step. 

You can wait for approval after each step, to know if you are going well.

For create-new-project, use these values

```bash
? Which certification does this belong to? python-v9
? What is the dashed name (in kebab-case) for this project? <>
? title: <>
? Choose a help category: Python
? Choose a block label: workshop
? Choose a block layout: challenge-grid
? What chapter should this project go in? python
? What module should this project go in? python-basics
? At which position: 6
```

I am unsure about the position, but it's really easy to change so don't worry too much about that, maybe 6 (It needs to end after the lecture on functions)
For the title maybe ask Naomi. The dashed name will be `workshop-<thing>` and the title `Build a <thing>`, for that thing you should see Naomi, she is really good with titles.

naomi-lgbt:
Title: Build a Kitchen Inventory Tracker

majestic-owl448:
dashed name becomes `workshop-kitchen-inventory-tracker`

PR #1085, opened 2025-12-18
Add strings workshop prototype for Python v9

RitamPal26:
Checklist:

Prototype for freeCodeCamp/freeCodeCamp#64624

## Description

This PR adds a prototype for the **strings workshop** in the Python v9/full stack certification. 

**Covers:**
- String concatenation, augmented assignment, and `str()` conversion
- F-strings for formatting
- String slicing (positive/negative indexing, step parameter)
- String methods: `.strip()`, `.lower()`, `.upper()`, `.title()`, `.replace()`, `.split()`, `.join()`, `.find()`, `.startswith()`, `.endswith()`, `.count()`
- Using `in` operator for validation

Ready for team review and feedback!

### DIFF
--- a/fullstack-cert/python-projects/strings-workshop-2/main.py
+++ b/fullstack-cert/python-projects/strings-workshop-2/main.py
@@ -23,9 +23,9 @@ print("Has .com: " + str(has_com))
 # PART 2: STRING METHODS - CASE CONVERSION
 # ========================================
 
-messy_name = "  SARAH DAVIS  "
+messy_name = "  sARaH dAVis  "
 
-messy_email = "  Sarah.Davis@COMPANY.COM  "
+messy_email = "  SARAH.DAVIS@COMPANY.COM  "
 
 clean_name = messy_name.strip().title()
 clean_email = messy_email.strip().lower()


jdwilkin4 (commented):
I read through the issue and looked through the PR and my initial thoughts were to break up this workshop into two workshops. 

Maybe one workshop for string concatenation, f strings and then a second workshop for the string methods.

But we can hear what the other curriculum team members think. So maybe it is fine to leave it all as one workshop.

Dario-DC:
I'm not against splitting it into two workshops. It really depends on how many steps do we want here. Probably it would be better to keep the initial workshops a bit shorter (?). If we agree we can create a separate issue for a second workshop.

jdwilkin4:
yeah, let's go with shortening this 👍🏾

RitamPal26:
Hey, so what are the next steps for me? Should I start working on them making one workshop for string concatenation, f strings and then a second workshop for the string methods, or something else?

majestic-owl448:
it's quite long, but it's all the same theme. And we have workshops that are quite enourmous.

If we want to divide in two workshops we can have this code divided in two different files (in two different folders) as two separate prototypes, and have a second contributor pick up the other half

jdwilkin4:
@RitamPal26 

We are still discussing the changes needed here. We should have a decision soon though on next steps for you 👍🏾 

@Dario-DC and @majestic-owl448 

> If we want to divide in two workshops we can have this code divided in two different files (in two different folders) as two separate prototypes, and have a second contributor pick up the other half

Yeah, that might be the better way to go. Especially since we have interest from contributors wanting to pickup more python work. 

We could have parts 1-4 be workshop 1 and parts 5 - end be workshop 2.
Then create another issue for workshop 2 and put it on Naomi's board.
thoughts?

majestic-owl448 (commented):
We can have a first workshop be parts 1 to 5, and a second one be the parts on methods 6-9, if you can divide them in two folders please

One thing that may need to be adjusted, considering that each part is best started with at least an empty terminal, deleting the `print`s from the code for the previous part, this allows to also have less noise in the terminal and be able to focus on what a small number of lines is doing.

 I would ask you reuse variables between sections, for example `employee_name` on line 35 could be instead the `full_name` created on line 10, in this way we can have less steps that just ask for creating variables, and have more steps dedicated to operators and methods and syntax

majestic-owl448:
@jdwilkin4 I see you have decided on the slicing to be in the other half, I'm fine either way

jdwilkin4:
parts 1 to 5 is fine with the first workshop 👍🏾

RitamPal26:
> We can have a first workshop be parts 1 to 5, and a second one be the parts on methods 6-9, if you can divide them in two folders please
> 
> One thing that may need to be adjusted,  I would ask you reuse variables between sections, for example `employee_name` on line 35 could be instead the `full_name` created on line 10, in this way we can have less steps that just ask for creating variables, and have more steps dedicated to operators and methods and syntax
> 
> I would ask you do this change considering that each part is best started with at least an empty terminal, deleting the `print`s from the code for the previous part, this allows to have less noise in the terminal and be able to focus on what a small number of lines is doing.
> 
> 

Yeah sure, I can do that

fullstack-cert/python-projects/strings-workshop-2/main.py
```diff
@@ -0,0 +1,83 @@
...
+
+clean_name = messy_name.strip().title()
+clean_email = messy_email.strip().lower()
+
+print("Original name: '" + messy_name + "'")
+print("Cleaned name: '" + clean_name + "'")
+print("Original email: '" + messy_email + "'")
+print("Cleaned email: '" + clean_email + "'")
+
+name_badge = clean_name.upper()
```
majestic-owl448, line 38: can you maybe make `messy_name` have a different capitalization than this? just as logic flow, so that it would not be less work to do `name_badge = messy_name.strip()`?

fullstack-cert/python-projects/strings-workshop-2/main.py
```diff
@@ -0,0 +1,83 @@
...
+print("Name badge: " + name_badge)
+
+# ========================================
+# PART 3: STRING METHODS - MANIPULATION
+# ========================================
+
+at_pos = clean_email.find("@")
+email_user = clean_email[:at_pos]
+display_name = email_user.replace(".", " ").title()
```
majestic-owl448, line 47: similar here. maybe the email can start all uppercase?

RitamPal26: Hey, thanks for pointing them out, I have made the requested changes.

majestic-owl448 (approved):
for me it's good, @Dario-DC will take an other look

Dario-DC (approved):
LGTM 🎉

Dario-DC:
I've just created a separate issue for the second workshop on string methods: https://github.com/freeCodeCamp/freeCodeCamp/issues/64869

Issue #1063, opened 2025-11-26
Add interfaces and types lab to main repo with tests


jdwilkin4:
Here is the prototype 

https://github.com/freeCodeCamp/CurriculumExpansion/pull/1035/files

I went ahead and approved it to help move the process a long so we can start adding it to the main repo since TypeScript is high priority. 

## Tasks left to do

- review the prototype and make any changes or refactors necessary
- Add the remaining user stories
- Add the project with user stories and tests to the main repo under the new frontend libraries v9 certification.

a2937:
I'm gonna to make a really quick change to the prototype. Then figure out what user stories are missing and what Typescript changes are necessary for it to run properly in the lab. 

Edit: I thought I was going to, but I'm not sure where we actually stand on naked motorcycles. To me, it sounded like a bad idea to include them but then they were in the dataset.

a2937:
There's like way too much CSS in the app. 900 lines just feels excessively excessive. It's probably going to be a good idea to trim a bunch of that.

jdwilkin4:
@a2937 

> Edit: I thought I was going to, but I'm not sure where we actually stand on naked motorcycles

I don't understand what this means.

a2937:
<img width="1467" height="692" alt="Screenshot of Code" src="https://github.com/user-attachments/assets/0c4abf25-abcb-42bd-823c-16c6b76e4b58" />

https://github.com/shootermv/CurriculumExpansion/blob/ae141e31ae2ea402d586bcb5ad9e9a2847c05597/fullstack-cert/typescript-projects/interface-and-types-lab/src/main.ts

It means I wanna know if the word "Naked" can appear anywhere in the curriculum; even in a non-explicit or non-suggestive context like this. 

It seems to be implied to be yes, as there is mention to it in both the JSON data and this lab.

majestic-owl448:
there are many naked things that are completely SFW, so I don't think that the word itself is an issue

a2937:
I think I know a thing that went wrong with the original user stories. It was too heavily focused on individual implementation details rather than the actual functionality.

PR #1088, opened 2025-12-19
feat: Add more complete user stories

a2937:
Checklist:

Related to https://github.com/freeCodeCamp/CurriculumExpansion/issues/1063 

The stories I added feel more like the kinds of the things that would be used for tests rather than as the user stories that would display at the top of the challenge. But campers also kind of need to know these things so that they can complete the challenge.

I also need to modify the solution code. I want to change the selectors from using `class` to using `id` since it seems more reliable to me.

### DIFF
--- a/fullstack-cert/typescript-projects/interface-and-types-lab/user-stories.md
+++ b/fullstack-cert/typescript-projects/interface-and-types-lab/user-stories.md
@@ -1,18 +1,31 @@
 # User Stories
 In this lab, you will build a MotoShop - electronic motorcycle market.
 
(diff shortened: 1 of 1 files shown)


jdwilkin4:
I don't think we need a PR to update the user stories. I am going to close this. 

You will be the one adding this to main. So you can just add the new user stories there.
It would be different other team members were going to work on this and need to look at the current user stories file. But in the case it is just you working on this.  

Same goes for the final solution. make any changes necessary to the solution code and add it to the PR in the main repo. The review team will test everything there.  👍🏾

PR #1090, opened 2025-12-20
fix: interfaces lab has function mentioned in user stories

a2937:
Checklist:

Related to https://github.com/freeCodeCamp/CurriculumExpansion/issues/1063

This is a very quick PR because it didn't actually exist when I started on the lab in the main repo.

### DIFF
--- a/fullstack-cert/typescript-projects/interface-and-types-lab/src/main.ts
+++ b/fullstack-cert/typescript-projects/interface-and-types-lab/src/main.ts
@@ -32,63 +32,44 @@ async function fetchMotorcycles() :  Promise<Motorcycle[]>
 }
 
 // MOTORCYCLE CARD COMPONENT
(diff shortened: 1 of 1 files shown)


majestic-owl448:
you can just update the final solution in the PR in the main repo and comment that you did that

jdwilkin4:
Yeah, I agree about adding the changes to the PR in the main repo instead of opening up a PR here.  

Opening PRs in the curriculum expansion for this particular case will just slow down the overall process. The TypeScript content is a high priority. So we don't want to slow down the process if it is not necessary.

If you are going to make changes to the user stories or final solution, you can just make them in the PR in the main repo. 

If you are planning on scrapping the project all together and creating something brand new, then I can see a reason for adding an extra review process with the curriculum expansion repo. But in this case, it is not necessary. 👍🏾

PR #1163, opened 2026-06-02
Prototype for javascript try/catch/finally/throw workshop

zenowinged:
### Checklist:

This PR introduces the working, minimalistic final prototype for the **Library Book Search using try/catch/finally/throw Debugging Workshop**.

This is a prototype for issue [#66931](https://github.com/freeCodeCamp/freeCodeCamp/issues/66931) in the freeCodeCamp repo.

### Notes
Although not in the final solution, I do have an additional step for the `debugger;` statement.

### DIFF
--- a/javascript-algorithms-and-data-structures/debugging-library-search/script.js
+++ b/javascript-algorithms-and-data-structures/debugging-library-search/script.js
@@ -1,0 +1,36 @@
+const library = [
+    "The Hobbit", "1984", "Dune", "Fahrenheit 451", "To Kill a Mockingbird",
+    "The Great Gatsby", "Moby Dick", "Pride and Prejudice", "Frankenstein"
+];
+
+const searchBtn = document.getElementById('search');
+const statusDisplay = document.getElementById('statusMessage');
+const bookInput = document.getElementById('bookTitleInput');
+const scanStatusDisplay = document.getElementById('scanStatus');
+
+const find_my_book = (title) => {
+    if (title === "") {
+        throw new Error("Search field cannot be empty");
+    }
+
+    if (!library.includes(title)) {
+        throw new Error(title + " not found!");
+    }
+
+    return title + " is available to borrow!";
+}
+
+searchBtn.addEventListener('click', () => {
+    try {
+        const result = find_my_book(bookInput.value);
+        if (result) {
+            statusDisplay.textContent = result;
+        }
+    } catch (err) {
+        statusDisplay.textContent = err.message;
+        console.error(err);
+    } finally {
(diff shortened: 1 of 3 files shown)


javascript-algorithms-and-data-structures/debugging-library-search/script.js
```diff
@@ -0,0 +1,36 @@
...
+    "The Hobbit", "1984", "Dune", "Fahrenheit 451", "To Kill a Mockingbird",
+    "The Great Gatsby", "Moby Dick", "Pride and Prejudice", "Frankenstein"
+];
+
+const searchBtn = document.getElementById('search');
+const statusDisplay = document.getElementById('statusMessage');
+const bookInput = document.getElementById('bookTitleInput');
+const scanStatusDisplay = document.getElementById('scanStatus');
+
+const find_my_book = (title) => {
```
jdwilkin4: workshops and lab use the camelCase naming convention instead of snake case. Also, I think `findBook` would be a better name.


```suggestion
const findBook = (title) => {
```

javascript-algorithms-and-data-structures/debugging-library-search/script.js
```diff
@@ -0,0 +1,36 @@
...
+    if (!library.includes(title)) {
+        throw new Error(title + " not found!");
+    }
+
+    return title + " is available to borrow!";
+}
+
+searchBtn.addEventListener('click', () => {
+    try {
+        const result = find_my_book(bookInput.value);
```
jdwilkin4: I would suggest adding trim as well. Otherwise, this would fail if you did something like adding a space at the end to a valid entry `1984 `

```suggestion
        const result = findBook(bookInput.value.trim());
```

javascript-algorithms-and-data-structures/debugging-library-search/script.js
```diff
@@ -0,0 +1,36 @@
...
+        if (result) {
+            statusDisplay.textContent = result;
+        }
```
jdwilkin4: this doesn't seem necessary since result will either throw an error or return a string that isn't empty. 

this should work 


```suggestion
       statusDisplay.textContent = result
```

zenowinged:
Got it! I'll apply the necessary changes. Thanks!

zenowinged:
hey @jdwilkin4 i've made the suggested changes please look into it whenever possible
thanks!

jdwilkin4 (approved):
Looks good. 

When you add this to the main repo, you will need to format your code using a formatter like prettier so the formatting is consistent 👍🏾

